PR #43, opened 2023-11-17
Add a Puma plugin for Solid Queue

djmb:
This allows you to run Solid Queue as part of a puma cluster, by adding

```
plugin :solid_queue
```

to your Puma config.

It will boot up the Solid Queue supervisor, leaving it to manage the scheduler and worker processes as normal.

If either Puma or the Solid Cache supervisor dies, the other will notice and shut itself down.

On Puma shutdown, it will signal Solid Queue to also shutdown.

cc @rosa

### DIFF
--- a/lib/puma/plugin/solid_queue.rb
+++ b/lib/puma/plugin/solid_queue.rb
@@ -1,0 +1,63 @@
+require "puma/plugin"
+
+Puma::Plugin.create do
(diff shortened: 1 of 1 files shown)


stillhart:
How do I control the puma solid_queue concurrency? The `config/solid_queue.yml` does not seem to apply here
```yaml
# config/solid_queue.yml
default: &default
  workers:
    - queues: "*"
      threads: 25
``

PR #64, opened 2023-12-19
Use ActiveRecord::Migration[7.0] in migrations

brunoprietog:
Hey,

First of all, thank you so much for amazing Solid Queue! ❤️

Migrations are using version 7.1, but support is provided from Rails 7.0 onwards, so users installing migrations using Rails 7.0 will get an error and will have to manually change the version of the migrations class. This PR fixes that.


rosa:
Oh, indeed! Great spot 🙏  Thank you!

PR #65, opened 2023-12-19
Update example for all queues sign

morgoth:
Without it, I'm getting error:
```
YAML syntax error occurred while parsing /path-to-app/config/solid_queue.yml. Please note that YAML must be consistently indented using spaces. Tabs are not allowed. Error: (<unknown>): did not find expected alphabetic or numeric character while scanning an alias at line 6 column 15
```
when starting `bundle exec rake solid_queue:start`

Please note that the latter example `staging*` works fine as is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ production:
     - polling_interval: 1
       batch_size: 500
   workers:
-    - queues: *
+    - queues: "*"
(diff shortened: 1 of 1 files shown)


rosa:
Oh, great spot! Thank you!

PR #69, opened 2023-12-19
Update README.md

jmarsh24:
There is a generator provided by the library but it's missing from the documentation. I have added the installation generator to the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,14 +39,24 @@ Or install it yourself as:
 $ gem install solid_queue
 ```
 
-Add the migration to your app and run it:
+Install Migrations and Set Up Active Job Adapter
+Now, you need to install the necessary migrations and configure the Active Job's adapter. Run the following commands:
+```bash
+$ bin/rails generate solid_queue:install
 ```
+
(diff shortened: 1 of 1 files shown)


brunoprietog:
I wonder if are also using Solid Queue in the tests and how do it if that is the case, since I see that the generator also sets it up in the test environment. I would think that would be used [ActiveJob::QueueAdapters::TestAdapter](https://api.rubyonrails.org/classes/ActiveJob/QueueAdapters/TestAdapter.html)

rosa:
Hey @jmarsh24, thanks for this! I'm not quite sure about the generator, in fact 😅 As @brunoprietog said, it's added to the `test` environment but we don't have yet a good way to start it and configure it for testing (even though it can be used for testing via an `inline` mode, but this is not quite fleshed out). I had wanted to keep the generator out of the first open-source release but I forgot to delete it. I'll do that instead.

rosa (approved):
Going to keep the generator in the end, after @excid3's #73, will just make a couple of changes to it. 

Thanks for this!

PR #73, opened 2023-12-19
Copy over example config on install

excid3:
The install steps eluded to a configuration not needed (an empty file will suffice) so I ran the install generator and ran into this error.

```bash
± bundle exec rails solid_queue:start
bin/rails aborted!
Configuration for Solid Queue not found in /Users/chris/code/solidqueue-example/config/solid_queue.yml

Tasks: TOP => solid_queue:start
(See full trace by running task with --trace)
```

This PR adds an example `config/solid_queue.yml` file to the generator to fix this error on install and provide an easy to customize example config.

Should make it a bit easier to use and customize. 👍 

Thanks for all your hard work @rosa!!

Edit: I just saw on your other PR that you were going to delete the generator, so feel free to close this if you want. This might be a useful reason to keep the generator?

### DIFF
--- a/lib/generators/solid_queue/install/install_generator.rb
+++ b/lib/generators/solid_queue/install/install_generator.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class SolidQueue::InstallGenerator < Rails::Generators::Base
+  source_root File.expand_path("templates", __dir__)
+
   class_option :skip_migrations, type: :boolean, default: nil, desc: "Skip migrations"
 
   def add_solid_queue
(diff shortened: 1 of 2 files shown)


excid3:
If the install generator is going away, would it be better to use a default config instead of raising an error on the missing config file?

kevinmirc:
Small suggestion: Since every field in `config/solid_queue.yml` is optional and falls back to the system defaults, I'd suggest having generator create a `config/solid_queue.yml` file with no contents. This would automatically use the system defaults and there is no concern of the template's values drifting from the hard-coded default values.

rosa:
Ahhh! @excid3 thanks a lot for pointing this out! This is totally my bad, I intended for Solid Queue to run without being configured just with the defaults, but I hadn't realised the file had to exist 🤦‍♀️ This was a last-minute change I did before open-sourcing the project 😅 Basically, what you said: 

> would it be better to use a default config instead of raising an error on the missing config file?

But also: 

> This might be a useful reason to keep the generator?

This is a good point! I think I'll keep the generator but will remove `development` and `test` environments from it, and will also make it so that you don't need to config file when starting out. I'll get this merged and will do all these changes on top of this. 

Thank you again! 🙏

rosa:
Fixed the issue with the file having to exist 😊

excid3:
Awesome, thanks @rosa!

PR #76, opened 2023-12-20
Don't raise if there's no config file in the default location

rosa:
Use the defaults in that case. Thanks to @excid3 for catching this in #73 🙌


excid3:
This will be much more beginner friendly. 🎉

PR #82, opened 2023-12-21
Make sure we wrap runnable code in app's executor.

nashby:
closes https://github.com/basecamp/solid_queue/issues/75


rosa (approved):
Nice catch! 🙏

PR #83, opened 2023-12-22
feature: add failed scope to job

adrianthedev:
I think it's useful and a common thing to query all the failed jobs.
I added a scope for it.

### DIFF
--- a/app/models/solid_queue/job/executable.rb
+++ b/app/models/solid_queue/job/executable.rb
@@ -17,6 +17,7 @@ module SolidQueue
         after_create :prepare_for_execution
 
         scope :finished, -> { where.not(finished_at: nil) }
+        scope :failed, -> { includes(:failed_execution).where.not(failed_execution: {id: nil}) }
       end
 
       %w[ ready claimed failed scheduled ].each do |status|


app/models/solid_queue/job/executable.rb
```diff
@@ -17,6 +17,7 @@ module Executable
         after_create :prepare_for_execution
 
         scope :finished, -> { where.not(finished_at: nil) }
+        scope :failed, -> { includes(:failed_execution).where.not(failed_execution: {id: nil}) }
```
rosa, line 20: Nice! 👏  In Mission Control, we go through `SolidQueue::FailedExecution` first to get the failed jobs. The reason is that the query is much faster because `solid_queue_failed_executions` table would be much smaller than the `solid_queue_jobs` table, but until we release that, this would be useful for sure 😊

adrianthedev: Got it. Makes sense!

I'm building a UI for Solid queue until Mission Control launches. 
I need to start the query from the `SolidQueue::Job` model and apply the query to it. I can't start from the `SolidQueue::FailedExecution` model.

When is Mission Control due to be out?

rosa: Nice! I hope it'll be before the end of January (being quite conservative and pessimistic). It's actually finished, just needs a bit of polishing and adding a couple of more things, so I hope much sooner than the end of January, but just to be on the safe side 😅

PR #90, opened 2023-12-25
Clean up executable.rb

### LINKED ISSUE #89: Method `failed_with` is defined repeatedly
https://github.com/basecamp/solid_queue/blob/282d75d9b6821b5e196db259a5f6bea43d6fe6d2/app/models/solid_queue/job/executable.rb#L53C11-L53C22

justinko:
Closes #89

### DIFF
--- a/app/models/solid_queue/job/executable.rb
+++ b/app/models/solid_queue/job/executable.rb
@@ -50,10 +50,6 @@ module SolidQueue
         finished_at.present?
       end
 
-      def failed_with(exception)
-        FailedExecution.create_or_find_by!(job_id: id, exception: exception)
-      end
-
       def discard
         destroy unless claimed?
       end


app/models/solid_queue/job/executable.rb
```diff
@@ -79,7 +75,6 @@ def ready
           ReadyExecution.create_or_find_by!(job_id: id)
         end
 
-
```
rosa, line 82: This new line was on purpose 😊 It adds a visual separation between methods related to each other, for enqueuing the job and a method for a setting not directly related to the 3 methods above.

justinko: Gotcha. Haven't seen that style before. I'll revert.

rosa (approved):
Thank you! 🙏

PR #101, opened 2023-12-29
Fix config.yml template indentation

allcentury:
Prior to this change, if all #'s are removed from the config template you're left with is a yaml file that solid_queue can't parse.

```
bundle exec rake solid_queue:start
rake aborted!
YAML syntax error occurred while parsing
/project/config/solid_queue.yml. Please note
that YAML must be consistently indented using spaces. Tabs are not
allowed. Error: (<unknown>): did not find expected key while parsing a
block mapping at line 1 column 1
```

reproducable in this repo via (remove comments first).  After fix:

```
ActiveSupport::ConfigurationFile.parse("./lib/generators/solid_queue/install/templates/config.yml")["development"]
=> {"batch_size"=>10, "dispatchers"=>[{"polling_interval"=>1,
"batch_size"=>500}], "workers"=>[{"queues"=>"*", "threads"=>5,
"processes"=>1, "polling_interval"=>0.1}]}
```

### DIFF
--- a/lib/generators/solid_queue/install/templates/config.yml
+++ b/lib/generators/solid_queue/install/templates/config.yml
@@ -1,4 +1,4 @@
-#default: &default
+# default: &default
 #   dispatchers:
 #     - polling_interval: 1
 #       batch_size: 500


rosa:
Oh! I can't reproduce this one. Is it possible you left the first line: 
```
#default: &default
```
commented-out?

allcentury:
I see this file:

```#default: &default
#   dispatchers:
#     - polling_interval: 1
#       batch_size: 500
#   workers:
#     - queues: "*"
#       threads: 5
#       processes: 1
#       polling_interval: 0.1
#
# development:
#  <<: *default
#
# test:
#  <<: *default
#
# production:
#  <<: *default
```

If I remove the `#` - I'm left with:

```yaml
default: &default
   dispatchers:
     - polling_interval: 1
       batch_size: 500
   workers:
     - queues: "*"
       threads: 5
       processes: 1
       polling_interval: 0.1

 development:
  <<: *default

 test:
  <<: *default

 production:
  <<: *default
```

Which fails with:

```
ActiveSupport::ConfigurationFile.parse("./lib/generators/solid_queue/install/templates/config.yml")
/Users/anthonyross/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/activesupport-7.1.0/lib/active_support/configuration_file.rb:29:in `rescue in parse': YAML syntax error occurred while parsing ./lib/generators/solid_queue/install/templates/config.yml. Please note that YAML must be consistently indented using spaces. Tabs are not allowed. Error: (<unknown>): did not find expected key while parsing a block mapping at line 1 column 1 (RuntimeError)
```

So that's problem #1.  


Problem #2 is that the `Rails.env` keys  (development, test, etc) still have a leading space when they should have zero.  

Problem #3 is that the default indentation has one extra space (3 instead of 2)

rosa:
Oh! I just realised the real issue is the missing space before `default`, in `#default`. My editor was handling that correctly via toggle comment shortcut, so I didn't see the problem, but I think the change should be this: 
```diff
diff --git a/lib/generators/solid_queue/install/templates/config.yml b/lib/generators/solid_queue/install/templates/config.yml
index 9c1856a..7270d0a 100644
--- a/lib/generators/solid_queue/install/templates/config.yml
+++ b/lib/generators/solid_queue/install/templates/config.yml
@@ -1,4 +1,4 @@
-#default: &default
+# default: &default
 #   dispatchers:
 #     - polling_interval: 1
 #       batch_size: 500
```

allcentury:
@rosa ya.  Looks like adding that one space back allows `ActiveSupport::ConfigurationFile.parse("./lib/generators/solid_queue/install/templates/config.yml")` to work. 

It still shows an extra leading space for me when I use https://github.com/tpope/vim-commentary but it's easy enough for me to remove.  Which patch do you prefer? 

Screenshot 2023-12-29 at 3 27 56 PM

rosa:
I prefer the second one; I think it looks nicer with the space after `#`.

rosa (approved):
Thank you so much!

PR #106, opened 2023-12-30
Ensure 60.seconds et al. works out-of-the-box

fractaledmind:
I was preparing a reproducible bug report for a separate issue, but couldn't get the script to run as I was always hit with a `NoMethodError` on the `60.seconds` call in the `mattr_accessor`.

### DIFF
--- a/lib/solid_queue.rb
+++ b/lib/solid_queue.rb
@@ -6,6 +6,9 @@ require "solid_queue/engine"
 require "active_job/queue_adapters/solid_queue_adapter"
 require "active_job/concurrency_controls"
 
+require "active_support/deprecator"
+require "active_support/core_ext/numeric/time"
+
 require "solid_queue/app_executor"
 require "solid_queue/processes/supervised"
 require "solid_queue/processes/registrable"


lib/solid_queue.rb
```diff
@@ -6,6 +6,9 @@
 require "active_job/queue_adapters/solid_queue_adapter"
 require "active_job/concurrency_controls"
 
+require "active_support/deprecator"
```
rosa, line 9: Why do you need this one?

fractaledmind: I hit an error in an IRB console trying to only `require "active_support/core_ext/numeric/time"`. However, I think I'll just close this for now, as this is really separate from my primary goal: to create a bug report script guide. I think I just need to figure out what all `require`s are necessary at the top of that script. And changes to the core lib are very likely not needed; just the right collection of `require`s at the top of that script

rosa: Thank you @fractaledmind! I'm planning to switch to [zeitwerk](https://github.com/fxn/zeitwerk) soon, I just haven't had the time yet, so hopefully you won't run into this again.

PR #112, opened 2024-01-02
git: Ignore folder for JetBrains IDEs

kiriakosv:
I'm not entirely sure if this is needed since someone can always ignore this folder on a local level. Feel free to close this PR if it's not needed.


rosa (approved):
Good idea! It definitely doesn't hurt 😊

PR #113, opened 2024-01-04
Exclude VS Code settings and .DS_Store file

anonychun:
- Ignored the Visual Studio Code settings folder to prevent unnecessary changes from being tracked.
- Exclude the system-specific .DS_Store file from version control.


rosa (approved):
👍  Thank you!

PR #126, opened 2024-01-15
Clear clearable jobs in a loop instead of batching

intrip:
ActiveRecord batching can be slow when looping over records filtered by non-PKs. We don't care about the order of deletion here, so we can just use a loop to delete the clearable jobs; this allows us to achieve better performance in general.

/cc @rosa @djmb

### DIFF
--- a/app/models/solid_queue/job/clearable.rb
+++ b/app/models/solid_queue/job/clearable.rb
@@ -11,7 +11,10 @@ module SolidQueue
 
       class_methods do
         def clear_finished_in_batches(batch_size: 500, finished_before: SolidQueue.clear_finished_jobs_after.ago)
-          clearable(finished_before: finished_before).in_batches(of: batch_size).delete_all
+          loop do
+            records_deleted = clearable(finished_before: finished_before).limit(batch_size).delete_all
+            break if records_deleted == 0
+          end
         end
       end
     end


app/models/solid_queue/job/clearable.rb
```diff
@@ -11,7 +11,14 @@ module Clearable
 
       class_methods do
         def clear_finished_in_batches(batch_size: 500, finished_before: SolidQueue.clear_finished_jobs_after.ago)
-          clearable(finished_before: finished_before).in_batches(of: batch_size).delete_all
+          loop do
+            records_deleted = \
+              SolidQueue::Job.
+                clearable(finished_before: finished_before).
```
intrip, line 17: `finished_before` is stable: This ensure that the loop will stop after processing al the records created before `finished_before`

app/models/solid_queue/job/clearable.rb
```diff
@@ -11,7 +11,10 @@ module Clearable
 
       class_methods do
         def clear_finished_in_batches(batch_size: 500, finished_before: SolidQueue.clear_finished_jobs_after.ago)
-          clearable(finished_before: finished_before).in_batches(of: batch_size).delete_all
+          loop do
+            records_deleted = clearable(finished_before: finished_before).limit(batch_size).delete_all
```
intrip, line 15: `finished_before` is stable: This ensures that we won't have infinite loops in fetching records.

rosa (approved):
Great! Thanks a lot @intrip 🙏

PR #134, opened 2024-01-23
Rebrand back to 37signals

northeastprince:
Also follows rails/rails#47467


rosa (approved):
Ha! I definitely copied these from elsewhere and didn't realise 😳  
Thank you!

Issue #120, opened 2024-01-10
Issue with "on_thread_error" override


chiraggshah:
I am trying to override `on_thread_error` by adding the following in application.rb.

```ruby
config.solid_queue.on_thread_error = ->(exception) {
  puts "*" * 100
  puts "exception: #{exception}"

  # do something

  Rails.error.report(exception, source: "solid_queue", severity: :error)
}
```

I have a sample job to trigger the above

```ruby
class SendMailJob < ApplicationJob
  def perform
    raise StandardError
  end
end
```

But the proc doesn't get executed when I run `SendMailJob.perform_later` in the rails console.

npezza93:
Doesn't seem link on_thread_error is ever called. If I `bundle open solid_queue` and throw a bunch of `puts "here"` into the default on_thread_error proc, nothing gets outputted to development.log

kylekeesling:
Likewise, I have the following config and it does not notify Honeybadger when an error occurs:

```
config.solid_queue.on_thread_error = ->(exception) {
  Honeybadger.notify(exception, context: {error_message: exception.message, source: "solid_queue"})
}
```

rosa:
Hey @chiraggshah, @npezza93, @kylekeesling thanks for reporting this and sorry for the delay. I've been working mostly on unrelated stuff this past week. 

`on_thread_error` wasn't quite intended for errors on the job itself, but rather errors in the thread that's executing the job, but around the job itself. For example, if you had an Active Record's thread pool too small for your number of threads and you got an error when trying to check out a new connection, `on_thread_error` would be called with that. 

For errors in the job itself, you could try to hook into Active Job's itself, similarly to what Sentry's client does [here](https://github.com/getsentry/sentry-ruby/blob/bd4f58a83042c29dd61102723012cfae2253024f/sentry-raven/lib/raven/integrations/rails/active_job.rb#L9-L32).

kylekeesling:
@rosa I started to wonder that based on the wording `thread` instead of `job`. Thanks for the clarification!

kylekeesling:
Using the Sentry example Rosa provided, I was able to get the following working in my app by adding it to `ApplicationJob`:

```ruby
class ApplicationJob < ActiveJob::Base
  perform do |job, block|
    capture_and_record_errors(job, block)
  end

  def capture_and_record_errors(job, block)
    Rails.error.set_context(**error_context(job))
    Rails.error.handle(Exception) { block.call }
  end

  def error_context(job)
    {
[...]

PR #139, opened 2024-01-29
Add section for exception handling

chiraggshah:
@rosa : I have added the documentation as discussed in #120. Let me know your feedback.


brunoprietog:
Why not use `rescue_from` instead? As stated in the [Active Job guide](https://guides.rubyonrails.org/active_job_basics.html#exceptions)

chiraggshah:
@brunoprietog : Thanks for the input. I have updated the PR.

brunoprietog:
Nice! Now it looks much cleaner

README.md
```diff
@@ -230,6 +233,32 @@ failed_execution.discard # This will delete the job from the system
...
+
+Note that, you will have to duplicate the above logic on `ActionMailer::MailDeliveryJob` too. That is because `ActionMailer` doesn't inherit from `ApplicationJob` but instead uses `ActionMailer::MailDeliveryJob` which inherits from `ActiveJob::Base`.
+
+```ruby
+# application_mailer.rb
+
+class ApplicationMailer < ActionMailer::Base
+  ActionMailer::MailDeliveryJob.rescue_from(Exception) do |exception|
+    Rails.error.report(exception)
+  end
```
rosa, line 259: With this suggestion and the above, the job will succeed because the exception will be rescued. I think you don't want this in the majority of cases, you want your job to fail so you can inspect it and retry it.

rosa (changes requested):
This looks good but I think we should re-raise the exception in both examples or people will have failing jobs succeed without an opportunity to examine or retry them.

rosa:
Also... I need to look a bit deeper into this, but I'm not sure using `around_perform` as it was used before, and `rescue_from` is fully equivalent. In the first case, we're scoping the block where an exception is captured to the `perform` method, and in the second case, we can rescue any exception even if it happens outside `perform` 🤔 I'm not sure if this makes a difference, I'd need to dig deeper into it, it's just that it's not completely obvious to me that these are equivalent.

README.md
chiraggshah: I have added the line to raise the exception.

chiraggshah:
@rosa : Good point. I too will have a look.

rosa (approved):

PR #143, opened 2024-02-06
Fix encoding error in sqlite

jpcamara:
When `Socket.gethostname` returns a hostname with a special character like apostrophe in `Basecamp’s-Computer`, an encoding error is raised `Encoding::UndefinedConversionError: "\xE2" from ASCII-8BIT to UTF-8`.

By forcing a utf-8 encoding, the error goes away and the process row properly inserts.

A similar issue occurred in the newrelic ruby gem, and required the same solution https://github.com/newrelic/newrelic-ruby-agent/commit/e1f65a830909adca5021f1d0f0c14b399579b7c8

### DIFF
--- a/lib/solid_queue/processes/registrable.rb
+++ b/lib/solid_queue/processes/registrable.rb
@@ -55,7 +55,7 @@ module SolidQueue::Processes
       end
 
       def hostname
-        @hostname ||= Socket.gethostname
+        @hostname ||= Socket.gethostname.force_encoding(Encoding::UTF_8)
       end
 
       def process_pid
--- a/test/models/solid_queue/process_test.rb
+++ b/test/models/solid_queue/process_test.rb
@@ -1,4 +1,5 @@
 require "test_helper"
+require "minitest/mock"
 
 class SolidQueue::ProcessTest < ActiveSupport::TestCase
   test "prune processes with expired heartbeats" do
@@ -11,6 +12,17 @@ class SolidQueue::ProcessTest < ActiveSupport::TestCase
 
     assert_difference -> { SolidQueue::Process.count }, -2 do
       SolidQueue::Process.prune
+    end
+  end
+
+  test "hostname's with special characters are properly loaded" do
+    worker = SolidQueue::Worker.new(queues: "*", threads: 3, polling_interval: 0.2)
+    hostname = "Basecamp’s-Computer"
+
+    Socket.stub :gethostname, hostname.force_encoding("ASCII-8BIT") do
+      worker.start
+      wait_for_registered_processes(1, timeout: 1.second)
+      assert_equal hostname, SolidQueue::Process.last.hostname
     end
   end
 end


jpcamara:
This also fixes an issue in MissionControl https://github.com/basecamp/mission_control-jobs/issues/61

lib/solid_queue/processes/registrable.rb
```diff
@@ -55,7 +55,7 @@ def kind
       end
 
       def hostname
-        @hostname ||= Socket.gethostname
+        @hostname ||= Socket.gethostname.force_encoding(Encoding::UTF_8)
```
rosa, line 58: I just have a question: why not try to convert the string to UTF-8 via `encode(Encoding::UTF_8)` rather than changing the encoding alone?

jpcamara: hmmm thanks for a good question! I think it's just a habit that i've typically reached for `force_encoding` in the past so I did it here. I guess most issues i've encountered have been encoding misinterpretations, vs needing to be re-encoded. 

But I think `encode` should work just as well, might work for more use-cases, and also doesn't mutate anything which is a nice side effect (or lack of a side effect 🙃 ). I'll change it

jpcamara: Huh, it actually doesn't work.

```rb
hostname = "Basecamp’s-Computer"
 => "Basecamp’s-Computer"
hostname.force_encoding("ASCII-8BIT")
 => "Basecamp\xE2\x80\x99s-Computer"
hostname.force_encoding("ASCII-8BIT").encode(Encoding::UTF_8)
(irb):3:in `encode': "\xE2" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
```

It also fails the same way if I use the direct result of `Socket.gethostname` on my machine:

 ```rb
Socket.gethostname.force_encoding("ASCII-8BIT") # works
hostname.force_encoding("ASCII-8BIT").encode(Encoding::UTF_8) # fails
# `encode': "\xE2" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
```

The return value of `gethostname` seems to always come back in binary format, but the underlying bytes, at least in my case, are already valid UTF-8 (`\xE2\x80\x99` is the byte sequence for "right single quotation mark"). So by forcing the encoding to be UTF-8 with `force_encoding`, it reinterprets the string correctly. 

When I use `encode` it seems to actually convert as if it was binary data to utf-8 and fails.

I don't know for certain forcing encoding couldn't cause an issue on some platform somewhere - but the same format has been in use by the newrelic ruby agent for the past 5 years (https://github.com/newrelic/newrelic-ruby-agent/commit/e1f65a830909adca5021f1d0f0c14b399579b7c8), so there's a good chance it's ok to keep as `force_encoding`.

rosa: Huh, how odd! The first case, I'd expect it to fail, but it's very strange the string comes from `Socket.gethostname` with the wrong encoding, I wonder why that's it... 

In any case, ok to keep `force_encoding` 👍

rosa (approved):

PR #149, opened 2024-02-06
Document database connection usage

matteeyah:
## Summary

Documents database connection pool usage in the README.

Solid Queue, by default uses 2 connections more than the what Rails apps are configured with by default. Documenting this can help minimize confusion.


matteeyah:
@rosa What do you think about this? I feel it's important enough to include it in the README.

rosa:
@matteeyah yeah, I think it makes sense to have this in the README, but not like this in its own section and advising to lower the number of threads (because in most cases you'd want to increase the number of connections instead). I'd like to have Solid Queue warn you when you have a too low number of connections, using 
https://github.com/basecamp/solid_queue/blob/7e8b0ec53fd8edfd19fa31da6b707b919fc7afab/lib/solid_queue/configuration.rb#L51-L54

So I'll close this one and will try to address the two together, warning and documenting.

PR #155, opened 2024-02-20
Add support for recurring tasks (cron style jobs)

### LINKED ISSUE #104: Implement cron-style, recurring tasks
A dispatcher would handle these, and the configuration will be included in `solid_queue.yml`.

rosa:
This PR introduces support for recurring (aka. cron-style) tasks. They can be included in the dispatcher's configuration as: 
```yml
  dispatchers:
    - polling_interval: 1
      batch_size: 500
      recurring_tasks:
        my_periodic_job:
          class: MyJob
          args: [ 42, { status: "custom_status" } ]
          schedule: every second
```
`recurring_tasks` is a hash/dictionary, and the key will be the task key internally. Each task needs to have a class, which will be the job class to enqueue, and a schedule. The schedule is parsed using [Fugit](https://github.com/floraison/fugit), so it accepts anything [that Fugit accepts as a cron](https://github.com/floraison/fugit?tab=readme-ov-file#fugitcron). You can also provide arguments to be passed to the job, as a single argument, a hash, or an array of arguments that can also include kwargs as the last element in the array. 

The job in the example configuration above will be enqueued every second as:
```ruby
MyJob.perform_later(42, status: "custom_status")
```

Tasks are enqueued at their corresponding times by the dispatcher that owns them, and each task schedules the next one. This is pretty much [inspired by what GoodJob does](https://github.com/bensheldon/good_job/blob/994ecff5323bf0337e10464841128fda100750e6/lib/good_job/cron_manager.rb).

It's possible to run multiple dispatchers with the same `recurring_tasks` configuration. To avoid enqueuing duplicate tasks at the same time, an entry in a new `solid_queue_recurring_executions` table is created in the same transaction as the job is enqueued. This table has a unique index on `task_key` and `run_at`, ensuring only one entry per task per time will be created. This only works if you have `preserve_finished_jobs` set to `true` (the default), and the guarantee applies as long as you keep the jobs around. 

Finally, it's possible to configure jobs that aren't handled by Solid Queue. That's it, you can a have a job like this in your app:
```ruby
class MyResqueJob < ApplicationJob
  self.queue_adapter = :resque

  def perform(arg)
    # ..
  end
end
```

You can still configure this in Solid Queue: 
```yml
  dispatchers:
    - recurring_tasks:
        my_periodic_resque_job:
          class: MyResqueJob
          args: 22
          schedule: "*/5 * * * *"
```
and the job will be enqueued via `perform_later` so it'll run in Resque. However, in this case we won't track any `solid_queue_recurring_execution` record for it and there won't be any guarantees that the job is enqueued only once each time. 

This pull request also introduces a new configuration option for the dispatcher, to opt-out of concurrency maintenance, via `concurrency_maintenance: false` (it's `true` by default). You can have multiple dispatchers and choose that some of them do concurrency maintenance but not all of them, as well as one/some of them being in charge of dispatching recurring tasks but not all of them.

Closes #104. 

Pending:
- [x] Update README with this new feature.


lib/solid_queue.rb
```diff
@@ -3,24 +3,16 @@
...
-require "solid_queue/processes/signals"
-require "solid_queue/configuration"
-require "solid_queue/pool"
-require "solid_queue/worker"
-require "solid_queue/dispatcher"
-require "solid_queue/supervisor"
+require "active_job"
+require "active_job/queue_adapters"
+
+require "zeitwerk"
```
fxn, line 9: ❤️

klenis:
I'm testing this branch with a frequent cart expiry job that runs every minute. My concern is the amount of noise generated in the jobs table. What are your thoughts on either having the option to add a condition to the task definition such as:

```
recurring_tasks:
  expire_carts_job:
    class: ExpireCartsJob
    schedule: every minute
    if: -> { Cart.candidates_for_expiry.any? }
```

or a task specific version of the `clear_finished_jobs_after` setting for automatic cleanup

rosa:
@klenis, this would depend on your job volume, but a job every minute would be 1,440 jobs per day and 10,080 jobs after one week. How would this compare to your current job volume? 

As a comparison, in HEY, the noise corresponding to recurring jobs is about ~2,000 per day, but that's negligible compared to regular jobs (over 10M / day). I think this might be the case for most users because the lowest time interval you can schedule jobs to run recurringly is 1 second. 

In case it helps, this is what we use on HEY to delete jobs that finished over 3 days ago:
```ruby
# config/application.rb

# Keep finished Solid Queue jobs for 3 days
config.solid_queue.clear_finished_jobs_after = 3.days
```
And then as part of our recurring tasks: 
```yaml
clear_solid_queue_finished_jobs:
  class: "CronJob"
  schedule: "42 * * * *"
  args: "SolidQueue::Job.clear_finished_in_batches(batch_size: 1000)"
```

Would something like this work for you?

klenis:
I guess the scheduled cleaner could be a viable solution. It would be nice to be able to pass `class_name:` to `clear_finished_in_batches` to simplify custom cleaning logic but I understand if you want minimise the public interface.

Thank you for taking the time to respond and great job with Solid Queue 👏

rosa:
> It would be nice to be able to pass `class_name: to clear_finished_in_batches` to simplify custom cleaning logic

Ohh, interesting idea! I hadn't thought about that as we didn't need that granularity when clearing jobs, but it's something I can certainly add 😊 Thank you!

weilandia:
> @klenis, this would depend on your job volume, but a job every minute would be 1,440 jobs per day and 10,080 jobs after one week. How would this compare to your current job volume? 
> 
> 
> 
> As a comparison, in HEY, the noise corresponding to recurring jobs is about ~2,000 per day, but that's negligible compared to regular jobs (over 10M / day). I think this might be the case for most users because the lowest time interval you can schedule jobs to run recurringly is 1 second. 
> 
> 
> 
> In case it helps, this is what we use on HEY to delete jobs that finished over 3 days ago:
> 
> ```ruby
> 
> # config/application.rb
> 
> 
> 
> # Keep finished Solid Queue jobs for 3 days
> 
> config.solid_queue.clear_finished_jobs_after = 3.days
> 
> ```
> 
> And then as part of our recurring tasks: 
> 
> ```yaml
> 
> clear_solid_queue_finished_jobs:
> 
>   class: "CronJob"
> 
>   schedule: "42 * * * *"
> 
>   args: "SolidQueue::Job.clear_finished_in_batches(batch_size: 1000)"
> 
> ```
> 
> 
> 
> Would something like this work for you? 
> 
> 
> 
> 

Hey @rosa 👋

Does BC run SolidQueue on a dedicated db?

rosa:
> Does BC run SolidQueue on a dedicated db?

Yes! We use it for HEY only (for now), and it has its own DB that shares the hardware with the app's main DB.

weilandia:
> 
> Yes! We use it for HEY only (for now), and it has its own DB that shares the hardware with the app's main DB. 


Thanks @rosa! 

We have similar job amounts/day--How big is your dedicated queue db?  Did y'all consider sharing your main db?

rosa:
> We have similar job amounts/day--How big is your dedicated queue db? Did y'all consider sharing your main db?

We did consider it, and in the beginning, when we started using Solid Queue in production, we were running it there (about ~1M jobs per day). We looked into how the write load would look like when moving all the jobs, compared it to the load from the application, and realised it'd be a little less than multiplying the existing write load by 2, leaving less margin for peaks. In the end, we decided to be cautious and moved it to its own DB, which shares the hardware with the main app's DB and other DBs, just a separate database, as we still had a lot of margin in terms of IOPS supported by our disks there, CPU and memory.

README.md
```diff
@@ -265,3 +267,48 @@ Solid Queue has been inspired by [resque](https://github.com/resque/resque) and
 
 ## License
```
brunoprietog, line 268: Shouldn't this be at the end?

n-at-han-k:
> > Does BC run SolidQueue on a dedicated db?
> 
> Yes! We use it for HEY only (for now), and it has its own DB that shares the hardware with the app's main DB.

Am I missing something in the solid_queue readme? Not quite sure how you'd set solid_queue up to use a separate database to the one that stores rails model data.

README.md
rosa: Oops, @brunoprietog thanks for spotting this! It should totally be at the end 😆

AristoRap:
Hi everyone, I was just testing out adding Solid Queue to my rails app, but I can't seem to get the recurring tasks to work.
I see some references to a solid_queue_recurring_executions table, but I don't see that in the migration or my Postgres db. Maybe I'm missing something?
I added the gem, installed and opted for directly setting `config.active_job.queue_adapter = :solid_queue` in my application.rb
The solid_queue.yml file contains the below:

default: &default
  dispatchers:
    - polling_interval: 1
      batch_size: 500
      recurring_tasks:
        my_frequent_job:
          class: MyFrequentJob
          schedule: every minute
  workers:
    - queues: "*"
      threads: 5
      processes: 1
      polling_interval: 0.1

development:
  <<: *default

test:
  <<: *default

production:
  <<: *default

I can trigger the jobs just fine manually, it's only the scheduled approach that doesn't seem to work for me.
If anyone has some suggestions, I'd appreciate some help! Thanks a lot in advance.

PR #157, opened 2024-02-21
Gemspec: Use https for GitHub URLs

packagethief:
Minor tweak to use https for GitHub URLs in the Gemspec 😅


rosa:
Oops, great catch, thank you so much @packagethief! 🙏 🙏

PR #166, opened 2024-03-06
Add option to limit clearing of jobs to specific class name(s)

klenis:
I'll give this a try. Please let me know if it could be improved.

This is a suggestion originating from https://github.com/basecamp/solid_queue/pull/155#issuecomment-1979577345 to add a `class_name:` option to `clear_finished_in_batches`.

Instead of exposing new configuration settings to Solid Queue for controlling the expiry of specific jobs this PR would enable anyone to write their own custom cleanup logic and still rely on the core cleanup functionality.

### DIFF
--- a/app/models/solid_queue/job/clearable.rb
+++ b/app/models/solid_queue/job/clearable.rb
@@ -6,13 +6,13 @@ module SolidQueue
       extend ActiveSupport::Concern
 
       included do
-        scope :clearable, ->(finished_before: SolidQueue.clear_finished_jobs_after.ago) { where.not(finished_at: nil).where(finished_at: ...finished_before) }
+        scope :clearable, ->(finished_before: SolidQueue.clear_finished_jobs_after.ago, class_name: nil) { where.not(finished_at: nil).where(finished_at: ...finished_before).where(class_name.present? ? { class_name: class_name } : {}) }
       end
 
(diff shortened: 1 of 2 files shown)


songjiz:
I think `SQ::Job.where(class_name: "xxxJob").clear_finished_in_batch` is better

klenis:
I still prefer that there is a public, documented and tested interface for basic cleanup scenarios. After giving this another thought, how about if Solid Queue supplied a clearing job instead? It would also play nice with the upcoming recurring task functionality. 

Something like this

```
class SolidQueue::ClearFinishedJobs < ApplicationJob
  def perform(batch_size: nil, finished_before: nil, queue_name: nil, class_name: nil, priority: nil)
    SolidQueue::Job.where(**({ queue_name:, class_name:, priority: }.compact)).clear_finished_in_batches(**({batch_size:, finished_before:}.compact))
  end
end
```

Using recurring tasks:

```
dispatchers:
  - recurring_tasks:
      clear_frequent_check_jobs:
        class: SolidQueue::ClearFinishedJobs
        args: [ { class_name: "FrequentCheckJob" } ]
        schedule: every hour
```

There is a note in the README that automatic clearing is coming and maybe this is not the right way forward. Let me know what you think. I can make a PR for it.

PR #170, opened 2024-03-08
Use defaults for MySQL

### LINKED ISSUE #169: Protocol mismatch; server version = 11, client version = 10 on rails main when running the tests
When I point to rails main, do `bundle update` then `rake` I get:

```
/Users/dorianmariefr/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/bundler/gems/rails-608c1bfe17a2/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb:35:in `rescue in new_client': Protocol mismatch; server version = 11, client version = 10 (ActiveRecord::ConnectionNotEstablished)
	from /Users/dorianmariefr/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/bundler/gems/rails-608c1bfe17a2/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb:24:in `new_client'
	from /Users/dorianmariefr/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/bundler/gems/rails-608c1bfe17a2/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb:145:in `connect'
	from /Users/dorianmariefr/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/bundler/gems/rails-608c1bfe17a2/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb:154:in `block in reconnect'
	from /Users/dorianmariefr/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/bundler/gems/rails-608c1bfe17a2/activesupport/lib/active_support/concurrency/null_lock.rb:9:in `synchronize'
```

dorianmariecom:
Fixes #169

### DIFF
--- a/test/dummy/config/database.yml
+++ b/test/dummy/config/database.yml
@@ -28,8 +28,6 @@ default: &default
   adapter: mysql2
   username: root
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  host: "127.0.0.1"
-  port: 33060
 <% end %>
 
 development:


test/dummy/config/database.yml
```diff
@@ -28,8 +28,6 @@ default: &default
   adapter: mysql2
   username: root
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  host: "127.0.0.1"
-  port: 33060
```
rosa, line 32: This won't work with the `mysql` service in `docker-compose.yml`: 
https://github.com/basecamp/solid_queue/blob/2a5256842b974c276e3dc0e8e83abb012351e37e/docker-compose.yml#L7-L13

dorianmariecom: using docker-compose sudenlly makes things work much better :D I still think the default mysql port should be used but docker-compose works

rosa: If you run `bin/setup`, it'll run `docker-compose` for you: 
```bash
docker-compose up -d --remove-orphans
docker-compose ps
```

The different port choice is so as not to interfere with any local MySQL server you might be running in the host. The container port is the default.

brendon: Mysql's X-Protocol uses `33060` so you'll likely have to try another one :D

https://dev.mysql.com/doc/mysql-port-reference/en/mysql-port-reference-tables.html

PR #185, opened 2024-03-23
Stop SolidQueue on puma restart when puma plugin is used.

### LINKED ISSUE #180: Address already in use error when using both tmp_restart & solid_queue puma plugin
Starting from a brand new Rails 7.1 app with postgres:

**config/puma.rb**

```ruby
plugin :tmp_restart
plugin :solid_queue
```

**app/controllers/home_controller.rb**

```ruby

class HomeController < ApplicationController
  def index
    FileUtils.touch(Rails.root.join('tmp/restart.txt'))
  end
end
```

**routes.rb**

```ruby
get '/', to: 'home#index'
```

When reaching the homepage, it should restart the server gracefully.
But instead I get:

```
* Restarting...
=> Booting Puma
=> Rails 7.1.3.2 application starting in development 
=> Run `bin/rails server --help` for more startup options
Puma starting in single mode...
* Puma version: 6.4.2 (ruby 3.3.0-p0) ("The Eagle of Durango")
*  Min threads: 5
*  Max threads: 5
*  Environment: development
*          PID: 20912
Exiting
/Users/max/.rvm/gems/ruby-3.3.0/gems/puma-6.4.2/lib/puma/binder.rb:334:in `initialize': Address already in use - bind(2) for "127.0.0.1" port 3000 (Errno::EADDRINUSE)
```

Then the process exists, and after this, it prints:

```
Detected Puma has gone away, stopping Solid Queue..
[SolidQueue] Terminating gracefully...
TRANSACTION (0.1ms)  BEGIN
  SolidQueue::Process Load (0.6ms)  SELECT "solid_queue_processes".* FROM "solid_queue_processes" WHERE "solid_queue_processes"."supervisor_id" = $1  [["supervisor_id", 6]]
  SolidQueue::Process Destroy (0.5ms)  DELETE FROM "solid_queue_processes" WHERE "solid_queue_processes"."id" = $1  [["id", 6]]
...
```

And the server is not running anymore. It's working without the `solid_queue` puma plugin. Am I missing something or does the plugin need to be improved to support restarting? Thanks a lot 🙂

nashby:
Without doing so SolidQueue would be running making Puma not properly stop so when Puma tries to start its process again we would get `Errno::EADDRINUSE` error.

Closes #180

### DIFF
--- a/test/integration/puma/plugin_test.rb
+++ b/test/integration/puma/plugin_test.rb
@@ -7,17 +7,18 @@ class PumaPluginTest < ActiveSupport::TestCase
   setup do
     FileUtils.mkdir_p Rails.root.join("tmp", "pids")
 
-    cmd = %w[
-      bundle exec puma
-        -b tcp://127.0.0.1:9222
-        -C test/dummy/config/puma.rb
-        --dir test/dummy
-        -s
-        config.ru
-    ]
+    Dir.chdir('test/dummy') do
+      cmd = %w[
+        bundle exec puma
+          -b tcp://127.0.0.1:9222
+          -C config/puma.rb
+          -s
+          config.ru
+      ]
 
-    @pid = fork do
-      exec(*cmd)
+      @pid = fork do
+        exec(*cmd)
+      end
     end
 
     wait_for_registered_processes(4, timeout: 3.second)
@@ -25,7 +26,7 @@ class PumaPluginTest < ActiveSupport::TestCase
 
   teardown do
     terminate_process(@pid, signal: :INT)
-    wait_for_registered_processes 0, timeout: 0.2.second
+    wait_for_registered_processes 0, timeout: 1.second
 
     JobResult.delete_all
   end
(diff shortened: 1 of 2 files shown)


test/integration/puma/plugin_test.rb
```diff
@@ -7,17 +7,18 @@ class PumaPluginTest < ActiveSupport::TestCase
...
-        -b tcp://127.0.0.1:9222
-        -C test/dummy/config/puma.rb
-        --dir test/dummy
-        -s
-        config.ru
-    ]
-
-    @pid = fork do
-      exec(*cmd)
+    Dir.chdir('test/dummy') do
```
nashby, line 10: I had to change the dir here this way instead of using `--dir` option because when you trigger puma's restart it can't find the app anymore since current dir is already `test/dummy` and it's trying to use `test/dummy` in addition.

rosa (approved):
Great spot, thank you! 🙏

PR #190, opened 2024-03-25
Fix scheduled scope

Intrepidd:
When looking on a way to compute a queue latency, I was looking for scopes to find jobs that have not run yet.

Looks like `scheduled` is the one, but the implementation seems incorrect. It's actually the same as `finished`.


rosa:
Ooh, totally! Great catch @Intrepidd 👍

PR #193, opened 2024-03-29
Used shorthand syntax for hash literals

chiraggshah:
Ruby 3.1 onwards, we can use shorthand syntax for hash literals. 
Ref: https://github.com/ruby/ruby/commit/c60dbcd1c55cd77a24c41d5e1a9555622be8b2b8


rosa:
Hey @chiraggshah, thanks! This syntax is not used elsewhere in this gem and I'm personally not a huge fan of it (yet 😅 ). I prefer to keep things consistent for now.

PR #195, opened 2024-04-01
Use beginless range for query instead of `< ?`

bensheldon:
This seems a little trivial to suggest (sorry!) but using a range in Rails 7.1 turns the query value into a proper bind parameter and preparable statement. This [will become unnecessary](https://github.com/rails/rails/pull/51139) in the next release of Rails. And I like the beginless/endless queries and all the other queries in SolidQueue that I could find use this style too 👍🏻 

Here's a writeup with a little more info: https://island94.org/2024/03/rails-active-record-will-it-bind


rosa:
Ooh, great catch! Thank you so much 🙏  And great article too 💯

Issue #137, opened 2024-01-29
Running out of connections


matteeyah:
## Summary

I've been getting a couple of connection errors in production. I have about 110-120 free connection on the database, but SolidQueue is reporting it couldn't get a connection.

Here's the error in question:

```
2024-01-29 14:28:07.568 [optonal_prod] worker.1 [SolidQueue] could not obtain a connection from the pool within 5.000 seconds (waited 5.081 seconds); all pooled connections were in use
```

Here's my `config/database.yml`

```
default: &default
  adapter: postgresql
  encoding: unicode
  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
  timeout: 5000

development:
  <<: *default
  database: optonal_development
  username: <%= 'postgres' if File.exist?('/.dockerenv') %>
  password:
  host: <%= 'db' if File.exist?('/.dockerenv') %>

test:
  <<: *default
  database: optonal_test
  username: <%= ENV.fetch('POSTGRES_USER', '') %>
  password: <%= ENV.fetch('POSTGRES_PASSWORD', '') %>
  host: <%= ENV.fetch('CI', '').empty? ? '' : 'postgres' %>

production:
  <<: *default
  database: optonal_production
```

Here's my `config/solidqueue.yml`

```
default: &default
   dispatchers:
     - polling_interval: 1
       batch_size: 500
   workers:
     - queues: [ real_time, background ]
       threads: 5
       processes: 1
       polling_interval: 0.1

development:
 <<: *default

test:
 <<: *default

production:
 <<: *default
```

## Proposal

Document which connection pool SolidQueue uses to connect to the database. Document how many connections does SolidQueue need (e.g. threads * processes).

matteeyah:
What's throwing me off here - I've set `threads` to 5, but I regularly see more than 5 entries in the `solid_queue_claimed_executions` table - does that mean it's trying to run more than 5 jobs at the same time? Because that would explain running out of connections since `config/database.yml` limits the number of connections in the pool to 5.

lrjbrual:
Hi @matteeyah, I am new using solid queue. 

When I push our codes with solid queue in the server, it seems I stock always with this: 

```
[2024-02-01 22:07:45] [SolidQueue] Starting Worker(pid=7, hostname=salesforce-connect-58dc4c96bc-qrj8w, metadata={:polling_interval=>2, :queues=>"*", :thread_pool_size=>2})
[...]

PR #201, opened 2024-04-06
Fix the worker count in the generated config file.

johnmcdowall:
The generated config file still has the worker count of 5, not 3. This led to issues in development as connections were quickly used up causing jobs to be stuck.

I think this was missed as part of #137


rosa:
Ahh, indeed! Good catch. Thank you! 🙏🏻

PR #208, opened 2024-04-12
Rely on `ActiveSupport::Notifications` and `ActiveSupport::LogSubscriber` for logging and instrumentation

### LINKED ISSUE #103: Switch to ActiveSupport::LogSubscriber
It looks like all logging is currently hardcoded. By switching to https://api.rubyonrails.org/classes/ActiveSupport/LogSubscriber.html instead (like Rails uses for most of its logging) it's easier for to change the log output - see eg https://github.com/roidrage/lograge and other Rails logging gems out there relying on this.

rosa:
This includes the basics to make it work, plus instrumentation for the most relevant actions. 

Closes #103.

### DIFF
--- a/lib/solid_queue/supervisor.rb
+++ b/lib/solid_queue/supervisor.rb
@@ -79,19 +79,24 @@ module SolidQueue
       end
 
       def graceful_termination
-        SolidQueue.logger.info("[SolidQueue] Terminating gracefully...")
-        term_forks
+        SolidQueue.instrument(:graceful_termination, supervisor_pid: ::Process.pid, supervised_pids: forks.keys) do |payload|
+          term_forks
 
-        wait_until(SolidQueue.shutdown_timeout, -> { all_forks_terminated? }) do
-          reap_terminated_forks
+          wait_until(SolidQueue.shutdown_timeout, -> { all_forks_terminated? }) do
+            reap_terminated_forks
+          end
+
+          unless all_forks_terminated?
+            payload[:shutdown_timeout_exceeded] = true
+            immediate_termination
+          end
         end
-
-        immediate_termination unless all_forks_terminated?
       end
 
       def immediate_termination
-        SolidQueue.logger.info("[SolidQueue] Terminating immediately...")
-        quit_forks
+        SolidQueue.instrument(:immediate_termination, supervisor_pid: ::Process.pid, supervised_pids: forks.keys) do
+          quit_forks
+        end
       end
 
       def term_forks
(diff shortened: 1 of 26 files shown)


lib/solid_queue/supervisor.rb
```diff
@@ -147,11 +152,11 @@ def reap_terminated_forks
...
-          start_fork(supervised_fork)
-        else
-          SolidQueue.logger.info "[SolidQueue] Tried to replace fork[#{pid}] (status: #{status.exitstatus}, fork[#{status.pid}]), but it had already died  (status: #{status.exitstatus})"
+        SolidQueue.instrument(:replace_fork, supervisor_pid: ::Process.pid, pid: pid, status: status) do |payload|
+          if supervised_fork = forks.delete(pid)
+            payload[:fork] = supervised_fork
+            start_fork(supervised_fork)
+          end
         end
       end
```
krschacht, line 161: @rosa This "Restarting fork" log line was a key clue in helping us debug and I know just how common it is for servers to auto-restart rails processes because of memory so what if we future-proof this for all the people in the years ahead who are going to be diagnosing auto-killed workers. :) 

In our situation the line previously read as:
```
[SolidQueue] Restarting fork[67] (status: )
```

You're already correcting the "restarting fork" language and I think the new log output for this situation will read as:
```
SolidQueue-#{ver} replace_fork (#{ms}) supervisor_pid: 12345, pid: 67, status:
```

That's slightly more helpful but for someone like me who isn't super familiar with SolidQueue it doesn't lead me much towards a possible root cause. I probably would have still posted an Issue to get help. What if you:

1. Notice if `status.pid` is nil and display `SIGKILL` for that
2. Make it clear that `status` is the status of the previous worker which is being replaced; I wouldn't infer that based on the new log output.
3. Include an additional `msg` key in the hash.

Here's my suggested improvement which incorporates these:
```
SolidQueue-#{ver} replace_fork (#{ms}) supervisor_pid: 12345, pid: 67, status: SIGKILL, msg: "Detected a terminated worker so it's being replaced with a new one. Exit status of previous worker was #{status || 'SIGKILL'}"
```

rosa: Thanks for taking the time to look into this one! I've added some more info in e5790d093748d2d181dfabc930cbf29174d441bd, which I hope it's useful in these cases.

krschacht: 👍 That looks great! Now there are definitely some clear clues as to what’s happening. I’m excited about all the investment in SolidQueue and am leaning in with a new product I’m building (rather than my old Sidekiq + Redis setup). Thanks for all your work on this!

PR #209, opened 2024-04-13
Added enqueue_after_transaction_commit setting

tannakartikey:
The `enqueue_after_transaction_commit` method was recently added to ActiveJob to be supported by the adapters.
PR: https://github.com/rails/rails/pull/51426

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,7 @@ There are several settings that control how Solid Queue works that you can set a
 - `preserve_finished_jobs`: whether to keep finished jobs in the `solid_queue_jobs` table—defaults to `true`.
 - `clear_finished_jobs_after`: period to keep finished jobs around, in case `preserve_finished_jobs` is true—defaults to 1 day. **Note:** Right now, there's no automatic cleanup of finished jobs. You'd need to do this by periodically invoking `SolidQueue::Job.clear_finished_in_batches`, but this will happen automatically in the near future.
 - `default_concurrency_control_period`: the value to be used as the default for the `duration` parameter in [concurrency controls](#concurrency-controls). It defaults to 3 minutes.
+- `enqueue_after_transaction_commit`: it defines weather the job should be enqueued after `after_commit`. [Read more](https://github.com/rails/rails/pull/51426) about it.
 
 
 ## Concurrency controls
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -171,6 +171,7 @@ There are several settings that control how Solid Queue works that you can set a
 - `preserve_finished_jobs`: whether to keep finished jobs in the `solid_queue_jobs` table—defaults to `true`.
 - `clear_finished_jobs_after`: period to keep finished jobs around, in case `preserve_finished_jobs` is true—defaults to 1 day. **Note:** Right now, there's no automatic cleanup of finished jobs. You'd need to do this by periodically invoking `SolidQueue::Job.clear_finished_in_batches`, but this will happen automatically in the near future.
 - `default_concurrency_control_period`: the value to be used as the default for the `duration` parameter in [concurrency controls](#concurrency-controls). It defaults to 3 minutes.
+- `enqueue_after_transaction_commit`: it defines weather the job should be enqueued after `after_commit`. [Read more](https://github.com/rails/rails/pull/51426) about it.
```
namolnad, line 174:
```suggestion
- `enqueue_after_transaction_commit`: it defines whether the job should be enqueued after `after_commit`. [Read more](https://github.com/rails/rails/pull/51426) about it.
```

namolnad (commented):
Just noticed one quick typo :)

rosa:
@tannakartikey thank you so much for this! It was in my to-do list so this is super helpful 😊 

I was thinking, however, about setting this by default to `true` 🤔 I think it's safer for most people and goes more in the spirit of this change in Active Job. 

If setting it to `true` by default, we'd need to modify [this section in the README](https://github.com/rails/solid_queue?tab=readme-ov-file#jobs-and-transactional-integrity) because the behaviour it warns about would be disabled now.

tannakartikey:
> @tannakartikey thank you so much for this! It was in my to-do list so this is super helpful 😊
> 
> I was thinking, however, about setting this by default to `true` 🤔 I think it's safer for most people and goes more in the spirit of this change in Active Job.
> 
> If setting it to `true` by default, we'd need to modify [this section in the README](https://github.com/rails/solid_queue?tab=readme-ov-file#jobs-and-transactional-integrity) because the behavior it warns about would be disabled now.

Glad to hear that, @rosa. Thank you for reviewing. 

I have set it to `true` by default and updated README. Let me know what you think. Thanks

zerobearing2:
Any idea when this might get merged? This issue has blocked us from staying on edge rails atm.

rosa:
Hey @zerobearing2, I'll merge now, sorry for the delay. I've been working on other stuff and travelling, and after the first review, I failed to get back to it to review the changes. 

> This issue has blocked us from staying on edge rails atm.

How so? Couldn't you just have set `enqueue_after_transaction_commit` to `always` or `never`? 🤔 

Anyway, I'll merge this now. Thanks so much again, @tannakartikey, the README change looks great.

zerobearing2:
> How so? Couldn't you just have set `config.active_job.enqueue_after_transaction_commit` to `always` or `never`? 🤔

We kept getting missing method error for solid queue adapter due to missing `enqueue_after_transaction_commit?` method. Wasn't aware setting `enqueue_after_transaction_commit` to `always` or `never` would skip the check. :thinking:  Thanks for merging though, cheers!

PR #175, opened 2024-03-11
support nil logger and ActiveSupport::Logger not Logger

dorianmariecom:
FIxes #168

PR #211, opened 2024-04-17
Handle nil logger situation

cardmagic:
There are times when ActiveRecord::Base.logger is set to nil and solid_queue should handle those times gracefully.


rosa:
Thanks @cardmagic! This is already handled by https://github.com/rails/solid_queue/pull/175, I need to merge that one.

Issue #144, opened 2024-02-06
migration path not respected when setting up solid_queue on separate database?


dianacamacho:
Firstly, thank you for all the work on solid queue so far! 

Apologies if this is lack of understanding and user-error on my part, but when the docs mention supporting solid queue on a different database than the main app (via `connects_to`), does that mean that the install/migration commands should automatically pick up on the settings I've defined in my app's `config/database.yml` for the given `connects_to` database key, or do I still need to manually tell solid queue which one to use when running those?

Currently, with the config and `database.yml` settings listed below for a separate `:solid_queue` db, when I run `bin/rails generate solid_queue:install` or `bin/rails solid_queue:install:migrations`, the migration files for the `SolidQueue` tables are copied to the `db/migrate` folder, resulting in the tables getting added to the `:primary` db instead of the `:solid_queue` one. However, I would have expected the install to copy them to `db/solid_queue_migrate`, to match the `migrations_paths` defined in `database.yml` for the `solid_queue` db)

```rb
# config/application.rb

config.active_job.queue_adapter = :solid_queue
config.solid_queue.connects_to = { database: { writing: :solid_queue, reading: :solid_queue } }
```
```yml
# config/database.yml

default: &default
  adapter: postgresql
  encoding: unicode
  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>

solid_queue: &solid_queue
  <<: *default
  migrations_paths: db/solid_queue_migrate

development:
  primary:
    <<: *default
    database: queue_demo_development
  solid_queue:
    <<: *solid_queue
    database: queue_demo_solid_queue_development

test:
  primary:
    <<: *default
    database: queue_demo_test

production:
  primary:
    <<: *default
  solid_queue:
    <<: *solid_queue
    url: <%= ENV["SOLID_QUEUE_DATABASE_URL"] %>
```

If I manually pass in the `DATABASE` as an arg to the migration command (`bin/rails solid_queue:install:migrations DATABASE=solid_queue`), then things do work as expected, with files copied to `db/solid_queue_migrate` and added to the `solid_queue` db). If this is how it's meant to work, maybe that can be made a little clearer in the docs. Thanks!

rosa:
@dianacamacho, thank you so much for this detailed and carefully written issue! You're spot on, `DATABASE=solid_queue` (or I think `--database solid_queue` works as well, but I haven't tested it) is how this should be used. 

Would you like to submit a PR to include this in the README?

[...]

PR #217, opened 2024-05-01
Support multiple databases for install generator

fractaledmind:
Currently, there is no way to specify that the migration files be installed into a separate database from the default. This is problematic especially for SQLite apps, since the recommended approach is to use a separate SQLite database for the queue. 

This change supports the existing behavior as well as specifying the `--database` option or the `--db` option.

### DIFF
--- a/lib/generators/solid_queue/install/install_generator.rb
+++ b/lib/generators/solid_queue/install/install_generator.rb
@@ -4,6 +4,7 @@ class SolidQueue::InstallGenerator < Rails::Generators::Base
   source_root File.expand_path("templates", __dir__)
 
   class_option :skip_migrations, type: :boolean, default: nil, desc: "Skip migrations"
+  class_option :database, type: :string, aliases: %i(--db), desc: "The database for your migration. By default, the current environment's primary database is used."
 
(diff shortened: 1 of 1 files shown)


fractaledmind:
@rosa: I can't figure out how the `solid_queue:install:migrations` Rake task is defined, so I can't then figure out if there is any way to specify a specific database for that task or not.

fractaledmind:
I will also add to the README and some tests

rosa:
Thank you so much @fractaledmind! This is related to https://github.com/rails/solid_queue/issues/144. 

The `solid_queue:install:migrations` task is [given by Rails's `Engine`](https://github.com/rails/rails/blob/087260d0fce3fa5aa50e8764dec14735e208c9e7/railties/lib/rails/engine.rb#L662-L675) and yes! It does allow specifying a database via the `DATABASE` environment variable, it was added in https://github.com/rails/rails/pull/48579. 

Thank you again! 🙏 🙇‍♀️

rosa:
In the end, I included this change in #308, as I had to change the generator. I completely missed that this one was completed; I thought something was missing! So sorry 😓

PR #219, opened 2024-05-03
Add test case to demonstrate that recurring tasks don't report exceptions to the error reporter

fractaledmind:
@rosa: I don't yet have a fix for this, but I'm hoping that a test case will help you to help me fix this.

### DIFF
--- a/test/dummy/config/solid_queue.yml
+++ b/test/dummy/config/solid_queue.yml
@@ -12,6 +12,10 @@ default: &default
           class: StoreResultJob
           args: [ 42, { status: "custom_status" } ]
           schedule: every second
+        periodic_raise_exception:
+          class: StoreResultJob
+          args: [ 42, { exception: RuntimeError } ]
+          schedule: every second
 
 development:
   <<: *default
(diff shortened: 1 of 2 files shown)


fractaledmind:
Also of note, if I run a `ErroringJob.perform_later` from the console, that exception is not reported either

rosa:
Sorry for totally ignoring this one! I don't know why I thought it was incomplete when you first opened it, so decided to wait until you completed it, and then completely forgot! I've come back to it now and I don't know why I thought it was incomplete 😳  It's not. 

This behaviour was initially intended. I noticed some error handling libraries [like `sentry-rails`](https://github.com/getsentry/sentry-ruby/blob/c3bcfa0401856d9ee7616c95ab9dfe988a89da30/sentry-rails/lib/sentry/rails/active_job.rb#L4) or [`rollbar-gem`](https://github.com/rollbar/rollbar-gem/blob/c4edf7e5094c3c043f2032fe560d217d5057c752/lib/rollbar/plugins/active_job.rb#L5-L17) that hook onto Active Job directly, so if I also reported this, we'd end up with duplicate error reports 🤔 But I'm not quite sure this is the right thing, perhaps it should be optional (via a new option, `on_job_error` like `on_thread_error`, or perhaps using `on_thread_error` for this)...  Or maybe it should just be documented, which is what #139 intended to do, but that one was incomplete in the end 🤔

rosa:
Ok, I went [with documenting this in the end](https://github.com/rails/solid_queue?tab=readme-ov-file#error-reporting-on-jobs), for now, but perhaps this should change in the future to a config option or something else 🤔

PR #231, opened 2024-05-22
Fixes #224.  I believe this is a Postgres only issue

### LINKED ISSUE #224: I think there is a race condition with limit_concurrency
Environment:
- Latest version of SolidQueue
- Postgresql@16.3
- MacOS

My workflow that seems to reliably cause the issue:
- I have a workflow where I create approximately 10  records in couple of seconds.  
- This model has an after_commit block that enqueues 3 jobs via perform_later.
- The after_commit code is wrapped in a transaction (the 3 jobs really should be enqueued all-or-nothing), but the SolidJob error was occurring with and without the transactional wrapper

One of the Jobs fairly consistently generates an error on 'INSERT INTO "solid_queue_semaphores".
For each of the block of 10 or so creates, the offending jobs solid_queue_semephore "key" value is JobName/User.id, and all 10 of the commits will have the same key. 

What I believe I'm seeing via the logs is:
* N jobs are being enqueued in parallel (looks like N of 2)
* Because of some unfortunate random timing, these 2 enqueue requests are both attempting to insert at effectively the same time.
* Because they have the same solid_queue_semephore "key" value and the "key" has a unique constraint, one of the enqueues gets rejected with a 'ActiveRecord::StatementInvalid (PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block' from deep within the bowels of Rails ActiveJob#enqueue.
* This is less than ideal ;-)

I haven't dug down into any of the SolidQueue code around enqueuing a job with limits_concurrency, but, at least with Postgres and it's version of MVCC, there is no way to preemptively detect a potential duplicate key prior to insert via a query.  It either has to be rescued or handled with an 'on conflict' clause (or the mySQL / sqlLite equivalent).

If my somewhat rushed conclusion is correct, then given limits_concurrency is a SolidQueue feature that enhances an enqueue request, it would be highly desirable for Solid to detect and retry internally.  Less desirable, but absolutely functional would be to raise an error that's specifically for this issue because the Job.perform_later is not the root cause, there isn't anything wrong with the database, you just got stepped on by combination of a database constraints and unfortunate timing, so the right answer is to resubmit job request it will likely work.

When I figure out how to get SolidQueue test environment working (I'm on a Mac and homebrew seems to require some kind of weird incantation to build a working version of mysql that works with docker), I'll see if I can develop a test for this specific case.

hms:
Per all of my earlier discussion, Postgres transactions become invalid upon any database contraint invalidating an operation and make all follow on database interactions invalid until a rollback is executed.

Semephore#attempt_creation uses a rather common database pattern to rely on database locks to reliably syncrhronize based on uniqueness of values.  The catch is that Postgresql requires a little more handholding to use this pattern.  By wrapping just the Semaphore.create statement in a SAVEPOINT (a pretend nested transaction), we have what appears to be a transparent operation that works across all of the other supported databases without any known consequences AND allows Postgres to continue to work as expected per the existing code.

This change works across the Solid supported databases (at least based on the tests) and I believe the extra overhead for the non-postgres databases is small enough that special casing the code or resorting to the complexity of dependency injection is just not worth it.

I added two tests to the concurrency_controls_test.rb. One to show the error is real and is easy to reproduce. The other to show the fix fixes it.

### DIFF
--- a/app/models/solid_queue/semaphore.rb
+++ b/app/models/solid_queue/semaphore.rb
@@ -66,11 +66,11 @@ module SolidQueue
         def check_limit_or_decrement = limit == 1 ? false : attempt_decrement
 
         def attempt_decrement
-          Semaphore.available.where(key: key).update_all(["value = value - 1, expires_at = ?", expires_at]) > 0
+          Semaphore.available.where(key: key).update_all([ "value = value - 1, expires_at = ?", expires_at ]) > 0
         end
 
         def attempt_increment
-          Semaphore.where(key: key, value: ...limit).update_all(["value = value + 1, expires_at = ?", expires_at]) > 0
+          Semaphore.where(key: key, value: ...limit).update_all([ "value = value + 1, expires_at = ?", expires_at ]) > 0
         end
 
         def key
--- a/app/models/solid_queue/semaphore.rb
+++ b/app/models/solid_queue/semaphore.rb
@@ -41,16 +41,30 @@ module SolidQueue
       end
 
       private
+
         attr_accessor :job
 
-        def attempt_creation
+        def attempt_creation_with_insert_on_conflict
+          results = Semaphore.insert({ key: key, value: limit - 1, expires_at: expires_at }, unique_by: :key)
+
+          if results.length.zero?
+            limit == 1 ? false : attempt_decrement
+          else
+            true
+          end
+        end
+
+        def attempt_creation_with_create_and_exception_handling
           Semaphore.create!(key: key, value: limit - 1, expires_at: expires_at)
           true
         rescue ActiveRecord::RecordNotUnique
-          if limit == 1 then false
-          else
-            attempt_decrement
-          end
+          limit == 1 ? false : attempt_decrement
+        end
+
+        if ActiveRecord::Base.connection.adapter_name == 'PostgreSQL'
+          alias attempt_creation attempt_creation_with_insert_on_conflict
+        else
+          alias attempt_creation attempt_creation_with_create_and_exception_handling
         end
 
         def attempt_decrement
--- a/app/models/solid_queue/semaphore.rb
+++ b/app/models/solid_queue/semaphore.rb
@@ -17,6 +17,17 @@ module SolidQueue
       def signal_all(jobs)
         Proxy.signal_all(jobs)
       end
+
+      # Requires a unique index on key
+      def create_unique_by(attributes)
+        if connection.supports_insert_conflict_target?
+          insert({ **attributes }, unique_by: :key).any?
+        else
+          create!(**attributes)
+        end
+      rescue ActiveRecord::RecordNotUnique
+        false
+      end
     end
 
     class Proxy
@@ -44,35 +55,22 @@ module SolidQueue
 
         attr_accessor :job
 
-        def attempt_creation_with_insert_on_conflict
-          results = Semaphore.insert({ key: key, value: limit - 1, expires_at: expires_at }, unique_by: :key)
-
-          if results.length.zero?
-            limit == 1 ? false : attempt_decrement
-          else
+        def attempt_creation
+          if Semaphore.create_unique_by(key: key, value: limit - 1, expires_at: expires_at)
             true
+          else
+            check_limit_or_decrement
           end
         end
 
-        def attempt_creation_with_create_and_exception_handling
-          Semaphore.create!(key: key, value: limit - 1, expires_at: expires_at)
-          true
-        rescue ActiveRecord::RecordNotUnique
-          limit == 1 ? false : attempt_decrement
-        end
-
-        if ActiveRecord::Base.connection.adapter_name == "PostgreSQL"
-          alias attempt_creation attempt_creation_with_insert_on_conflict
-        else
-          alias attempt_creation attempt_creation_with_create_and_exception_handling
-        end
+        def check_limit_or_decrement = limit == 1 ? false : attempt_decrement
 
         def attempt_decrement
-          Semaphore.available.where(key: key).update_all([ "value = value - 1, expires_at = ?", expires_at ]) > 0
+          Semaphore.available.where(key: key).update_all(["value = value - 1, expires_at = ?", expires_at]) > 0
         end
 
         def attempt_increment
-          Semaphore.where(key: key, value: ...limit).update_all([ "value = value + 1, expires_at = ?", expires_at ]) > 0
+          Semaphore.where(key: key, value: ...limit).update_all(["value = value + 1, expires_at = ?", expires_at]) > 0
         end
 
         def key
--- a/app/models/solid_queue/semaphore.rb
+++ b/app/models/solid_queue/semaphore.rb
@@ -61,7 +61,7 @@ module SolidQueue
           limit == 1 ? false : attempt_decrement
         end
 
-        if ActiveRecord::Base.connection.adapter_name == 'PostgreSQL'
+        if ActiveRecord::Base.connection.adapter_name == "PostgreSQL"
           alias attempt_creation attempt_creation_with_insert_on_conflict
         else
           alias attempt_creation attempt_creation_with_create_and_exception_handling
--- a/test/integration/concurrency_controls_test.rb
+++ b/test/integration/concurrency_controls_test.rb
@@ -83,7 +83,7 @@ class ConcurrencyControlsTest < ActiveSupport::TestCase
 
     # C would have started in the beginning, seeing the status empty, and would finish after
     # all other jobs, so it'll do the last update with only itself
-    assert_stored_sequence(@result, [ "C" ])
+    assert_stored_sequence(@result, ["C"])
   end
 
   test "run several jobs over the same record sequentially, with some of them failing" do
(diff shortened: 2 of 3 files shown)


hms:
@rosa 

I'm pausing on this for just a little as I did a little more digging (probably should have checked before coding a fix....) and there are some significant concerns around adding sub-transactions / savepoint for high volume deployments, especially those with replication / failover.

While my proposed solution is database agnostic (ignoring the Postgres performance issues referenced above), the tests cannot be.  Both SqLite and mySQL work without a testable failure with either my PR or the original SolidQueue implementation.

There look to be some non-nested transaction / savepoint implementations, but they all require different syntax for each of the 3 supported databases.  I'm open to investigating those alternatives, but it would help to understand if the SolidQueue team would be open to a PR based on database specific syntax and how to organize the code for any PRs with this approach.

Open issues that I could use some direction on:

- Does the SolidQueue team consider this a SolidQueue problem, a Rails upstream issue, or somewhere else I'm not looking at yet?
  - If the SolidQueue tam would prefer to defer to the upstream work that appears to be under consideration, is the team open to a temporary solution for the problem that does exist now?
  - If yes to a SolidQueue fix now, can I get some direction on the team's requirements for how to structure / organize / call out database specific implementations where and when necessary?
  
Thanks,

Hal

hms:
@rosa 

I've revamped this PR to avoid nested transaction.

 Postgres, unlike mysql and sqlite, invalidates the current transaction on
    a database contraint violation such as an attempt to insert
    a duplicate key into a unique index.
    
    SolidQueue uses a pretty standard design pattern of application level
    conflict detection and resolution via error handling as part of it's
    concurrency management in Semaphore::Proxy#attempt create (via create! and rescue).
    
    For Postgres based implementations when:
    * enqueing a job inside a transaction
    * under load such that the race condition triggerng the simultanious conflicted insert
    
    the transaction would become unrecoverable at the application level.
    
    There are two simple paths to address this issue in Postgres:
    * Nested transactions
    * Insert using 'on conflict do nothing'
    
    Nested transaction are the easiest to implement and are portable across all 3
    of the Rails standard databases. Unfortunately, nested transactions have a
    known and very signifant performance problem, especially for databases under
    load and/or replicated databases with long running queries.
    
    Each of the big three database inplements the ANSI standard Insert on conflict do nothing
    in a very different way.
    
    Because of this, this PR only replaces the current implementation of
    application level conflict handling for Postgres.  It takes advantage
    of the Rails standard ActiveRecord::Persistence::Insert method supporting on
    conflict semantics.

rosa:
Hey @hms, so sorry for the delay here! I've been totally focused on other stuff (and other projects) but I'm finally trying to wrap up everything for v1.0 and would like to get this fix. This last approach looks great to me. I have some suggestions for the implementation, but it's possible that, being so long ago, you no longer have time to implement these. In that case just let me know, happy to add these myself.

hms:
@rosa 

I'd love any feedback / suggestions to make the PR better and more durable for the project.  I'm happy to get this wrapped up to releaseable standards. 

Hal

test/integration/concurrency_controls_test.rb
```diff
@@ -183,7 +183,46 @@ class ConcurrencyControlsTest < ActiveSupport::TestCase
     assert job.reload.ready?
   end
 
+  if ActiveRecord::Base.connection.adapter_name == "PostgreSQL"
```
rosa, line 186: Since we have `TARGET_DB`, I think I'd use that here, like [here](https://github.com/rails/solid_queue/blob/f0ca6b43b5cefc97b40e3154be6ee535d77b8450/test/models/solid_queue/job_test.rb#L257-L267).

test/integration/concurrency_controls_test.rb
```diff
@@ -183,7 +183,46 @@ class ConcurrencyControlsTest < ActiveSupport::TestCase
...
+      assert SolidQueue::Semaphore.count == 0
+
+      SolidQueue::Semaphore.create!(key: key, value: limit - 1, expires_at: expires_at)
+      assert SolidQueue::Semaphore.count == 1
+
+      assert_raises ActiveRecord::RecordNotUnique do
+        SolidQueue::Semaphore.create!(key: key, value: limit - 1, expires_at: expires_at)
+      end
+    end
+  end
```
rosa, line 208: Hmm... actually, I'm not completely sure if perhaps we need this test at all, since it seems it's testing more something internal of Active Record, rather than something we have in Solid Queue 🤔

app/models/solid_queue/semaphore.rb
```diff
@@ -41,16 +41,30 @@ def signal
       end
 
       private
+
```
rosa, line 44: In this project I'm following the style of not having a space after `private` when the class has public and private methods, so this one was on purpose 😅

app/models/solid_queue/semaphore.rb
```diff
@@ -41,16 +41,30 @@ def signal
...
           true
         rescue ActiveRecord::RecordNotUnique
-          if limit == 1 then false
-          else
-            attempt_decrement
-          end
+          limit == 1 ? false : attempt_decrement
+        end
+
+        if ActiveRecord::Base.connection.adapter_name == "PostgreSQL"
```
rosa, line 64: It seems there's a method that checks directly whether the adapter supports `unique_by` exposed in [Active Record's Abstract adapter](https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/AbstractAdapter.html#method-i-supports_insert_conflict_target-3F), that we could use simply as 
```ruby
connection.supports_insert_conflict_target?
```
I think this might be more robust than checking the adapter name, because other adapters for PostgreSQL might have a different name, like [the PostGIS adapter](https://github.com/rgeo/activerecord-postgis-adapter).

app/models/solid_queue/semaphore.rb
```diff
@@ -41,16 +41,30 @@ def signal
...
+          alias attempt_creation attempt_creation_with_insert_on_conflict
+        else
+          alias attempt_creation attempt_creation_with_create_and_exception_handling
```
rosa, line 67: I'm thinking that this all could perhaps live in the `Semaphore` class in a simpler way. Something like this method: 
```diff
diff --git a/app/models/solid_queue/semaphore.rb b/app/models/solid_queue/semaphore.rb
index 34bf8dc..1448e67 100644
--- a/app/models/solid_queue/semaphore.rb
+++ b/app/models/solid_queue/semaphore.rb
@@ -17,6 +17,17 @@ module SolidQueue
       def signal_all(jobs)
         Proxy.signal_all(jobs)
       end
+
+
+      def create_unique_by(attributes)
+        if connection.supports_insert_conflict_target?
+          insert(attributes, unique_by: :key).any?
+        else
+          create!(attributes)
+        end
+      rescue ActiveRecord::RecordNotUnique
+        false
+      end
     end

     class Proxy
```

rosa: Then that could be used in `Proxy` class like this (BTW I might remove that class in the future and leave everything in `Semaphore`... it was not to have to pass all the job attributes around, but still... anyway, not the goal of this PR): 
```ruby
        def attempt_creation
          if Semaphore.create_unique_by(key: key, value: limit - 1, expires_at: expires_at)
            true
          else
            check_limit_or_decrement
          end
        end

        def check_limit_or_decrement
          if limit == 1 then false
          else
            attempt_decrement
          end
        end
```
And you wouldn't need to define aliases.

test/integration/concurrency_controls_test.rb
hms: That is better.   I'll submit the change.

test/integration/concurrency_controls_test.rb
hms: While this really is more of a test of AR, its value is that we have a confirmation of outcomes with the two different ways to add records to the Semaphore table and we show the two mechanisms "side-by-side".  I realized that it was missing an assert to meet that goal.  See what you think with the extra line.  

In the end, if you still feel strongly about cleaning this test up a little, I'll remove it.

app/models/solid_queue/semaphore.rb
hms: So funny and just goes to show a bit of tunnel vision.  I spent a fair amount of time digging around for an alternative to the original nested transactions methodology and in that digging I came across this method in the Rails adaptors.  

The tunnel vision part was no thinking to use those methods as the much, much better way to test which of the code paths to take.

Thanks for this -- it's a great reminder to slow down and think before coding.

hms:
@rosa 

Thank you for your feedback and suggestions. The PR is significantly better because of them.

I have a PR ready.  How do you feel about rebasing and squishing the older commits that, given the last set of changes, don't add much value from a history perspective?  The only downside is I think your comments might get lost in that process and given how  important they are to the finished PR, I'm not sure how I feel about that.

Hal

rosa:
Hey @hms! Thanks a lot for that! 🙏 No need to squash the commits, totally as you prefer. Feel free to push things and keep the original history.

test/integration/concurrency_controls_test.rb
rosa: Sounds good! I don't feel strongly about this one 😊

hms:
@rosa I could use some help here.  The tests run clean in my local environment over multiple runs.  

Interestingly, at least part of the error output refers to the Postgres transactional error that was the root cause for this PR.  I'm a RSpec person, so I'm not nearly as familiar with mini-test as I should be.  Does it wrap tests in transactions?  If so, then any test that hits a database managed constraint (unique, checks, foreign keys, etc) are all going to be at risk for trigging: PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block (ActiveRecord::StatementInvalid)

Hal

rosa:
Ohhh yes! I think some of those failures are legit and not related to your changes. I've got a similar fix to what you did here in the works in https://github.com/rails/solid_queue/pull/272/commits/3d40b1a8f11e35bcdb7d77b96db23a7dde1a7654

rosa:
Thanks again for this, @hms, amazing work 🙏 🙇

Issue #232, opened 2024-05-27
Fugit version is 1.11.0 now


ukolovda:
Latest Fugit version is 1.11.0 (minor changes).

rosa:
Would you like to open a PR to update?

PR #234, opened 2024-05-28
Update fugit version to 1.11.0

ukolovda:
Addressed #232


ukolovda:
@rosa  how do you think, it will be better use 
```ruby
spec.add_dependency "fugit", "~> 1.11"
```
for version condition?

rosa (approved):
Thank you!

PR #258, opened 2024-07-13
Fix Supervisor monitoring in Puma's plugin

### LINKED ISSUE #222: `monitor_solid_queue` not working on Puma Plugin
Hi🖐️
I'm using the puma plugin to start a solid_queue supervisor.
To make sure that puma is monitoring the supervisor, I kill the supervisor process, but the puma process does not terminate.
Possible causes,
In the solid_queue puma plugin, the registration to the background is done in the hook of the boot event, and [puma executes the function registered in the background before boot](https://github.com/puma/puma/blob/3169cf607ae3978eabf22c325f02eaefc8ca5c45/lib/puma/single.rb#L46), so I think that `monitor_solid_queue` is not being executed.
https://github.com/rails/solid_queue/blob/1d41b530048f3d3ccdbe83cdbe4c8cbe3543dd32/lib/puma/plugin/solid_queue.rb#L17

If the registration to the background was done outside of the boot event hook, `monitor_solid_queue` seemed to be executed.
```
  def start(launcher)
    @log_writer = launcher.log_writer
    @puma_pid = $$

    launcher.events.on_booted do
      @solid_queue_pid = fork do
        Thread.new { monitor_puma }
        SolidQueue::Supervisor.start(mode: :all)
      end
    end

    launcher.events.on_stopped { stop_solid_queue }
    launcher.events.on_restart { stop_solid_queue }

    in_background do
      monitor_solid_queue
    end
  end
```

Solid Queue: 0.3.0
Ruby: 3.2.2 
Puma: 6.4.2

Thanks

kolbasov:
Puma fires up the background hooks before the ```on_booted``` event. Since the plugin registers the ```monitor_solid_queue``` after Puma is fully booted, it doesn't get executed.

The fix is to move the ```monitor_solid_queue``` out of the ```on_booted``` and check whether the SolidQueue has started in the ```solid_queue_dead?```.

Fixes #222 

/cc @K0H205

### DIFF
--- a/lib/puma/plugin/solid_queue.rb
+++ b/lib/puma/plugin/solid_queue.rb
@@ -7,14 +7,14 @@ Puma::Plugin.create do
     @log_writer = launcher.log_writer
     @puma_pid = $$
 
+    in_background do
+      monitor_solid_queue
+    end
+
     launcher.events.on_booted do
       @solid_queue_pid = fork do
         Thread.new { monitor_puma }
         SolidQueue::Supervisor.start(mode: :all)
-      end
-
-      in_background do
-        monitor_solid_queue
       end
     end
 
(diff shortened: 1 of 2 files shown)


rosa (approved):
Sorry for the delay, @kolbasov. Thanks a lot for doing this! 🙏

kolbasov:
Most welcome, @rosa!

PR #269, opened 2024-07-29
Implement an `async` execution mode to run Solid Queue workers and dispatchers in the same process

rosa:
This PR builds on @bensheldon's initial spike in #115, to run workers and dispatchers in the same process as the supervisor instead of running them as forks. It also considerably refactors the way the supervisor works and it's started, to make room for this. 

For now, running in this new `async` mode is only supported via the Puma plugin but I plan to allow it also when running Solid Queue's supervisor alone in the near future, as a separate PR. 

This PR also introduces a breaking change in how Solid Queue is started. Prior this change, you could choose whether you wanted to run workers, dispatchers or both, by starting Solid Queue as `solid_queue:work` or `solid_queue:dispatch`. With this PR, the only option available is `solid_queue:start`. Whether the supervisor starts workers, dispatchers or both will depend on your configuration. For example, if you don't configure any dispatchers, only workers will be started. That is, with this configuration:

```yml
production:
  workers:
    - queues: [ real_time, background ]
      threads: 5
      polling_interval: 0.1
      processes: 3
```
the supervisor will run 3 workers, each one with 5 threads, and no supervisors.


rosa:
We're now running this branch on HEY to ensure it still works well in the regular forked mode.

PR #293, opened 2024-08-22
Fixed broken rollback migratio in processes' name

tmimura39:
This is a migration fix added by https://github.com/rails/solid_queue/commit/76d2c0f62e60663407bf0bba384e8d667f9cc3e3.
Since `up` is a change to NOT NULL, `down` must undo it.


tmimura39:
Hey @rosa, I did not report the `GitHub issue` as I think it was a simple mistake.
This is not that big a problem, but it is a source of confusion and has been corrected.

PR #308, opened 2024-08-28
Add a `bin/jobs` binstub to run the supervisor more easily

rosa:
This also fixes that the supervisor couldn't be started in `async` mode unless using the Puma plugin, and other minor issues in the installation like avoiding overwriting an existing `config/solid_queue.yml` file and accepting a `--database` parameter to specify a separate Solid Queue DB to install the migrations. The new `bin/jobs` binstub is copied over as part of the installation.


floehopper:
Is the new `bin/jobs` bin stub equivalent to `bundle exec rake solid_queue:start`? Is either one preferred over the other?

rosa:
Great question! It's mostly equivalent, witih a few differences. It allows passing a `--config_file` option with a different location for the workers and dispatchers configuration (with the Rake task you need to set an environment variable, `SOLID_QUEUE_CONFIG`, if you wanted to change that). It also behaves a bit differently with respect to eager loading, I believe, because Rake tasks ignore `config.eager_load = true`, there's an additional setting `config.rake_eager_load` to get this enabled, but I think many people might not have it enabled. 

I'd say `bin/jobs` is preferred, to be consistent with other Rails commands/tasks that have their own binstub like this.

floehopper:
@rosa Thanks for explaining!

Issue #311, opened 2024-08-29
Support for instrumenting polling?


gjtorikian:
Hello! I was wondering if you'd accept a PR that implements instrumentation around polling. Something pretty much like this:

```ruby
def with_polling_volume(&block)
  SolidQueue.instrument(:polling) do
    if SolidQueue.silence_polling? && ActiveRecord::Base.logger
      ActiveRecord::Base.logger.silence(&block)
    else
      yield
    end
  end
end
```

The reason being that even though `silence_polling` silences log statements, it does not silence instrumentation. That means that if the Rails app is hooked up to (for example) OpenTelemetry, the monitoring system receives thousands of `BEGIN` / `COMMIT` statements signifying nothing more than the polling is being checked. Can you tell when SQ was deployed? :P 

![graf](https://github.com/user-attachments/assets/b2289b02-705c-4745-9ba9-33262f529535)

Currently, I have a monkey patch to support ignoring this:

```ruby
module SolidQueue
  module Processes
    module PollerExtension
      def with_polling_volume
        OpenTelemetry::Instrumentation::PG.with_attributes("polling.solid_queue" => "true") do
          super
        end
      end
    end
  end
end

require "solid_queue/processes/poller_extension"

SolidQueue::Processes::Poller.prepend(SolidQueue::Processes::PollerExtension)
```

I can then instruct my collector to just filter out any `polling.solid_queue` events. This works, but it would be nice to remove my hack. 😆

gjtorikian:
No worries!

gjtorikian:
If anyone in the future comes here, the solution is now:

```ruby
ActiveSupport::Notifications.subscribe("polling.solid_queue") do
  OpenTelemetry::Trace.current_span.add_attributes({
    "polling.solid_queue" => "true",
  })
end
```

I hope your world is nice.

timwis:
@gjtorikian Would you mind clarifying where you put that snippet? When I try it, there is no current_span. Wondering if perhaps it's just outdated? I'm also getting a ton of noise in my tracing system.

gjtorikian:
I placed it into `config/initializers/solid_queue.rb`. [`current_span`](https://github.com/open-telemetry/opentelemetry-ruby/blob/4834e0056e877019ef266a852ba24f96e36f194f/api/lib/opentelemetry/trace.rb#L48-L51) is also still present in `main` for the OTel gem.

[...]

PR #312, opened 2024-08-29
Wrap polling within instrumentation

gjtorikian:
Should close https://github.com/rails/solid_queue/issues/311. I wasn't sure how interesting the tests should be, but I did my best to show _something_ happening.

### DIFF
--- a/test/integration/instrumentation_test.rb
+++ b/test/integration/instrumentation_test.rb
@@ -37,6 +37,21 @@ class InstrumentationTest < ActiveSupport::TestCase
     assert_event events.first, "claim", process_id: process.id, job_ids: jobs.map(&:id), claimed_job_ids: jobs.map(&:id), size: 3
   end
 
+  test "polling emits events" do
+    3.times { StoreResultJob.perform_later(42) }
+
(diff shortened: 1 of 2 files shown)


test/integration/instrumentation_test.rb
```diff
@@ -37,6 +37,25 @@ class InstrumentationTest < ActiveSupport::TestCase
...
+
+      wait_while_with_timeout!(3.seconds) { SolidQueue::ReadyExecution.any? }
+      process = SolidQueue::Process.last
+
+      worker.stop
+      wait_for_registered_processes(0, timeout: 3.second)
+    end
+
+    assert_equal 5, events.size
+    assert_event events.last, "polling"
```
rosa, line 56: Nice! I think you can remove a few variables here, like `process`, `jobs`, `worker`, to make this test simpler, since we don't need them to check anything.

gjtorikian: ok! I removed as many as I could to get a minimum running example. I think `process` still needs to be there!

rosa (approved):
Nice! Thank you so much! 🙏

PR #321, opened 2024-09-04
Slim down installer and have it use a final schema instead of migrations

dhh:
Following the same pattern we took with Solid Cache. This entails collapsing all the migrations into a premade queue_schema.rb and switching the default configuration to use a database configuration under the `queue` key. This configuration can now be set in config/solid_queue.yml, just like it's possible to do with Solid Cache. Drop all mentions and focus on running SQ off the same database.


spiritualpad:
If we would like to keep SolidQueue tables on a separate database server from the main app, how would we do that now?

dhh:
That's now the default. The schema is independent, you point solid queue to the database you want to use with `config.solid_queue.connects_to = { database: { writing: :your_db_name } }` and match that in `config/database.yml`.

PR #325, opened 2024-09-05
Update rake task name

andyatkinson:
There's no:
bin/rails solid_queue:install

in the current version, but presumably this task was changed to:

bin/rails solid_queue:install:migrations

which is a current task name

```sh
bin/rails -T solid_queue
bin/rails solid_queue:dispatch            # start solid_queue dispatcher to enqueue scheduled jobs
bin/rails solid_queue:install:migrations  # Copy migrations from solid_queue to application
bin/rails solid_queue:start               # start solid_queue supervisor to dispatch and process jobs
bin/rails solid_queue:work                # start solid_queue supervisor to process jobs
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Solid Queue can be used with SQL databases such as MySQL, PostgreSQL or SQLite, 
 Solid Queue is configured by default in new Rails 8 applications. But if you're running an earlier version, you can add it manually following these steps:
 
 1. `bundle add solid_queue`
-2. `bin/rails solid_queue:install`
+2. `bin/rails solid_queue:install:migrations`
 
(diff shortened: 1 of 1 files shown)


dhh:
https://github.com/rails/solid_queue/blob/main/lib/solid_queue/tasks.rb#L3

andyatkinson:
Makes sense. I don't know why `bundle add solid_queue` initially added this old version constraint `"~> 0.1.2"`. I removed that and ran `bundle update solid_queue` to get `0.8.1`, which has those updated tasks:

```sh
bin/rails -T solid_queue
bin/rails solid_queue:install  # Install Solid Queue
bin/rails solid_queue:start    # start solid_queue supervisor to dispatch and process jobs
```
😊

Issue #328, opened 2024-09-06
Same database for queue


lethaale:
Hi @rosa!

I want to deploy a single database with both normal data and solid queue data and I can't find in the README how.

Can you help? Thanks in advance!

dhh:
Take the contents of the queue_schema.rb file, shove it into a migration, and remove the "config.solid_queue.connects_to" line from config/environments/production.rb 👍

lethaale:
Thanks @dhh! 
Same goes for Solid Cache, I assume? 
Do you want me do a PR and add it to README?

rosa:
> Same goes for Solid Cache, I assume?

Yes!

> Do you want me do a PR and add it to README?

That'd be super nice if you have time 🙏 😊

PR #334, opened 2024-09-08
Add instructions to run with a single database

lethaale:
https://github.com/rails/solid_queue/issues/328

When running a small / hobby project (maybe to test solid queue) it's good to be able to test it with one Database. I propose to add instructions to the README.

If this PR gets accepted, happy to create a very similar one for Solid Cache


rosa (approved):
Looks great, thank you so much! 🙏

PR #335, opened 2024-09-08
Fix typo

anonychun:
fix typo from solid_cache to solid_queue


rosa (approved):
Oops, great catch! 🙏  Thank you!

PR #339, opened 2024-09-11
add hook for job

qinmingyuan:
For some reasons,  I really want this feature:
1. add callback after job finished;
2. add more scope for job model; 

If add this hook, I can override job model more easily.

### DIFF
--- a/app/models/solid_queue/job.rb
+++ b/app/models/solid_queue/job.rb
@@ -66,3 +66,5 @@ module SolidQueue
     end
   end
 end
+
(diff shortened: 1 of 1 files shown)


rosa:
Hey @qinmingyuan, thanks for opening this PR! 

Would you be able to do this using existing [Active Job's hooks](https://guides.rubyonrails.org/active_job_basics.html#callbacks)? For example, `after_perform`? I'm wary of introducing Solid Queue-specific features when this is something that feels more on the Active Job's purview 🤔

qinmingyuan:
Ok,  I will close this PR.

PR #345, opened 2024-09-13
Update Thor CLI options to respect ENV variables for default values

toydestroyer:
With the introduction of `bin/jobs` in #308, I noticed that it ignores the `SOLID_QUEUE_CONFIG` environment variable (and the recently introduced `SOLID_QUEUE_RECURRING_SCHEDULE`), while `rails solid_queue:start` still respects both of them.

This PR is my attempt to fix it.

Please let me know if it is worth adding tests for the `SolidQueue::Cli`.

Also, I hope that the direction is correct, although the logic duplicates `default_options` from `SolidQueue::Configuration`.

### DIFF
--- a/lib/solid_queue/cli.rb
+++ b/lib/solid_queue/cli.rb
@@ -5,13 +5,11 @@ require "thor"
 module SolidQueue
   class Cli < Thor
     class_option :config_file, type: :string, aliases: "-c",
-      default: Configuration::DEFAULT_CONFIG_FILE_PATH,
-      desc: "Path to config file",
+      desc: "Path to config file (default: #{Configuration::DEFAULT_CONFIG_FILE_PATH}).",
       banner: "SOLID_QUEUE_CONFIG"
 
     class_option :recurring_schedule_file, type: :string,
-      default: Configuration::DEFAULT_RECURRING_SCHEDULE_FILE_PATH,
-      desc: "Path to recurring schedule definition",
+      desc: "Path to recurring schedule definition (default: #{Configuration::DEFAULT_RECURRING_SCHEDULE_FILE_PATH}).",
       banner: "SOLID_QUEUE_RECURRING_SCHEDULE"
 
     class_option :skip_recurring, type: :boolean, default: false,


rosa:
Oops, thank you! I had broken this, then fixed it, then broken it again 🤦🏻‍♀️

lib/solid_queue/cli.rb
```diff
@@ -5,12 +5,12 @@
 module SolidQueue
   class Cli < Thor
     class_option :config_file, type: :string, aliases: "-c",
-      default: Configuration::DEFAULT_CONFIG_FILE_PATH,
+      default: ENV["SOLID_QUEUE_CONFIG"] || Configuration::DEFAULT_CONFIG_FILE_PATH,
```
rosa, line 8: I was thinking that perhaps the easiest way would be not to provide any defaults here so that `Configuration` is the only place where we need to care about defaults 🤔

toydestroyer: That works, but the downside is that it no longer appears in `bin/jobs --help`.

Before:
```
bin/jobs -h
Commands:
  jobs help [COMMAND]  # Describe available commands or one specific command
  jobs start           # Starts Solid Queue supervisor to dispatch and perform enqueued jobs. Default command.

Options:
  -c, [--config-file=SOLID_QUEUE_CONFIG]                          # Path to config file
                                                                  # Default: config/queue.local.yml
      [--recurring-schedule-file=SOLID_QUEUE_RECURRING_SCHEDULE]  # Path to recurring schedule definition
                                                                  # Default: config/recurring.local.yml
      [--skip-recurring]                                          
                                                                  # Default: false
```

After
```
bin/jobs -h                                                                            
Commands:
  jobs help [COMMAND]  # Describe available commands or one specific command
  jobs start           # Starts Solid Queue supervisor to dispatch and perform enqueued jobs. Default command.

Options:
  -c, [--config-file=SOLID_QUEUE_CONFIG]                          # Path to config file
      [--recurring-schedule-file=SOLID_QUEUE_RECURRING_SCHEDULE]  # Path to recurring schedule definition
      [--skip-recurring]                                          
                                                                  # Default: false
```

Let me know if I should proceed with the changes.

rosa: Ah, good point! Maybe we could just add this to the description? It's also that it's a bit misleading that the value of the env var is a default... 🤔  because if you unset that, the default will change 🤔 

I'm thinking of something like this: 
```diff
diff --git a/lib/solid_queue/cli.rb b/lib/solid_queue/cli.rb
index d36ae5f..947a86c 100644
--- a/lib/solid_queue/cli.rb
+++ b/lib/solid_queue/cli.rb
@@ -10,8 +10,7 @@ module SolidQueue
       banner: "SOLID_QUEUE_CONFIG"

     class_option :recurring_schedule_file, type: :string,
-      default: Configuration::DEFAULT_RECURRING_SCHEDULE_FILE_PATH,
-      desc: "Path to recurring schedule definition",
+      desc: "Path to recurring schedule definition (default: #{Configuration::DEFAULT_RECURRING_SCHEDULE_FILE_PATH}).",
       banner: "SOLID_QUEUE_RECURRING_SCHEDULE"
```

toydestroyer: Done:

```
bin/jobs -h
Commands:
  jobs help [COMMAND]  # Describe available commands or one specific command
  jobs start           # Starts Solid Queue supervisor to dispatch and perform enqueued jobs. Default command.

Options:
  -c, [--config-file=SOLID_QUEUE_CONFIG]                          # Path to config file (default: config/queue.yml).
      [--recurring-schedule-file=SOLID_QUEUE_RECURRING_SCHEDULE]  # Path to recurring schedule definition (default: config/recurring.yml).
      [--skip-recurring]                                          
                                                                  # Default: false
```

rosa (approved):
Nice! Thank you so much! 🙏

PR #354, opened 2024-09-18
Fix: Replace static table names in schema definition

m2meier:
The schema definition in `db/queue_schema.rb` uses static plural table names. If rails runs with `config.active_record.pluralize_table_names = false`, this leads to an error as soon as solid_queue tries to access the DB.

The static strings for the table names in the schema definition are replaced by the `table_name` method of the corresponding model class. This ensures that the tables are named correctly in plural or singular respectively.

### DIFF
--- a/lib/generators/solid_queue/install/templates/db/queue_schema.rb
+++ b/lib/generators/solid_queue/install/templates/db/queue_schema.rb
@@ -1,5 +1,5 @@
 ActiveRecord::Schema[7.1].define(version: 1) do
-  create_table "solid_queue_blocked_executions", force: :cascade do |t|
+  create_table SolidQueue::BlockedExecution.table_name, force: :cascade do |t|
     t.bigint "job_id", null: false
     t.string "queue_name", null: false
     t.integer "priority", default: 0, null: false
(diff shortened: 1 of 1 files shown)


rosa:
Huh, I didn't know about `config.active_record.pluralize_table_names`! 

@m2meier, do you run into the same problems for Active Storage, Action Text, Action Mailbox...?

m2meier:
> @m2meier, do you run into the same problems for Active Storage, Action Text, Action Mailbox...?

Same problem for ActiveStorage. My pull request: https://github.com/rails/rails/pull/52970. On ActionText and ActionMailbox I'm currently investigating.

rosa:
I wonder why this started failing after Rails 7.1, though 😕  Did Rails handle this automatically in the past? I see the `pluralize_table_names` config option has existed for way longer. Maybe this is a regression of some sort?

m2meier:
> I wonder why this started failing after Rails 7.1, though 😕 Did Rails handle this automatically in the past? I see the `pluralize_table_names` config option has existed for way longer. Maybe this is a regression of some sort?

The option already existed when I started with Rails (v5 or earlier). But before the introduction of ActiveStorage, there were no tables that Rails created itself. And only since v7.1 does `pluralize_table_names` lead to a problem with these tables.

rosa:
> And only since v7.1 does `pluralize_table_names` lead to a problem with these tables.

Right! So, this must be a regression of some sort. I think that's what should be addressed instead of working around it here or in the other core frameworks 🤔

PR #355, opened 2024-09-18
Add missing example for `config.solid_queue.connects_to`

miguno:
The `connects_to` parameter description in the current README has an incomplete sentence that ends with "For example:".

This PR adds such an example to the README.


rosa:
Oops, thank you! I removed the example by mistake solving some conflicts. Let me see what we had before... I might just restore that instead.

rosa:
Ahhh, no, I had removed that config option from that section in https://github.com/rails/solid_queue/commit/305534e23b0b08df19b897d07b87eb52f48f0eaf, then re-introduced it by mistake when fixing conflicts. I'll just remove it again.

rosa:
Ok, removed in https://github.com/rails/solid_queue/commit/2ed2cfe7a80846956857ffefa9cd7dd1e3d8b03e, so I'm going to close this one. Thanks a lot for caching this!

PR #375, opened 2024-10-07
Add compatibility with strict_loading_by_default

### LINKED ISSUE #122: Not compatible with strict loading by default
It looks like some of the queries aren't compatible with a `strict_loading_by_default` configuration.

With this configuration in `application.rb`, `solid_queue` doesn't work.

```
config.active_record.strict_loading_by_default = true
```

I will try to find some time to dig into his and open a fix.

JoeDupuis:
Fixes #122

Allows the app to run in `strict_loading_by_default` mode by explicitly turning off strict loading on all Solid Queue models.

Also enables `strict_loading_by_default` on the dummy app to prevent strict loading issues from being introduced.

##### Alternative / strict_loading enabled
I tried to support strict_loading instead of turning it off, but after spending time on it, I don't think it's worth it. It is a little bit faster, but bring a bunch of complexity. 
I opened a draft PR on my fork to demonstrate what it would look like: https://github.com/JoeDupuis/solid_queue/pull/1
Even if we decide to go down the road of enabling strict_loading in Solid Queue, we should merge this PR first to unblock strict_loading apps while working on the fix.

### DIFF
--- a/app/models/solid_queue/record.rb
+++ b/app/models/solid_queue/record.rb
@@ -3,6 +3,7 @@
 module SolidQueue
   class Record < ActiveRecord::Base
     self.abstract_class = true
+    self.strict_loading_by_default = false
 
     include DistinctValues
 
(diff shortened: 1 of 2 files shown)


rosa:
Sorry for taking so long to merge this one! I wasn't sure about it but I just tweaked it a bit and I think this is fine. Thank you!

JoeDupuis:
No worries! Thank you!

PR #381, opened 2024-10-13
Bump .ruby_version to latest Ruby version (3.3.5)

berkos:
👋  Thanks for the review in advance. 🙏 

This PR matches locally the ruby version with the [CI ](https://github.com/rails/solid_queue/blob/main/.github/workflows/main.yml#L25).
Also corrects a couple of minor typos spotted while going through the Readme.


README.md
```diff
@@ -252,7 +252,7 @@ class MyJob < ApplicationJob
 
   # ...
 ```
-- `key` is the only required parameter, and it can be a symbol, a string or a proc that receives the job arguments as parameters and will be used to identify the jobs that need to be limited together. If the proc returns an Active Record record, the key will be built from its class name and `id`.
+- `key` is the only required parameter, and it can be a symbol, a string or a proc that receives the job arguments as parameters and will be used to identify the jobs that need to be limited together. If the proc returns an ActiveRecord record, the key will be built from its class name and `id`.
```
rosa, line 255: I think [Active Record is written separate](https://guides.rubyonrails.org/active_record_basics.html) 🤔 Unless it's the specific Ruby module, but in that case I think I'd have written `ActiveRecord::Base`.

rosa (changes requested):
Thanks a lot and sorry for the delay!

README.md
berkos: Thanks for the review, that makes sense. I wrongly thought it was referred to the module. Corrected.

rosa (approved):
Thank you so much! 🙏

Issue #329, opened 2024-09-06
Running `rails db:migrate` clears out `queue_schema.rb`.


gregschmit:
I am trying to install latest solid queue into a project I'm working on, but when I run `rails db:migrate` then it clears out the `queue_schema.rb` file. I can load the schema with `rails db:prepare`, however I do have to run migrations in my primary database, but it always clears out the queue schema. Is there a migration file or multiple migration files for solid queue anymore?

gregschmit:
I guess I should just take the schema and dump it into a migration file. Sorry if this is a noobie question but I haven't used multiple DBs in Rails before so I'm trying to figure out if my configuration is broken or if this is expected behavior.

gregschmit:
Maybe if this is the normal current behavior solid queue install should put a migration into the `solid_queue_migrate` path? Because it does say to set the database configuration `migrations_paths` to `queue_migrate`, so I think that would indicate something should be there.

rosa:
Huh, sorry for the trouble, @gregschmit! That seems really strange! It should just work without you having to do anything, and it shouldn't clear the queue schema. Would you mind sharing your `database.yml` configuration?

gregschmit:
Sure, here is my database.yml:

```
default: &sqlite
  adapter: sqlite3
  pool: <%= ENV.fetch("RAILS_MAX_THREADS", 10) %>
  timeout: 5000

development:
  primary:
    <<: *sqlite
    database: db/development.sqlite3
  solid_queue:
    <<: *sqlite
    database: db/development_queue.sqlite3
    migrations_paths: db/solid_queue_migrate

test:
  primary:
    <<: *sqlite
    database: db/test.sqlite3
  solid_queue:
    <<: *sqlite
    database: db/test_queue.sqlite3
    migrations_paths: db/solid_queue_migrate

production:
  primary:
    <<: *sqlite
    database: db/production.sqlite3
  solid_queue:
    <<: *sqlite
    database: db/production_queue.sqlite3
    migrations_paths: db/solid_queue_migrate
```

In hindsight to me it makes perfect sense that the schema would be cleared, because there are no migration files inside `db/solid_queue_migrate`. By the way, I originally had everything set to `queue_migrate` and `queue_schema.rb` because that was the default but ended up prepending `solid_` just for consistency. I was able to make everything work by putting the template `queue_schema.rb` into a migration file inside `the `db/queue_migrate` directory and now things are smooth.

gregschmit:
[...]

PR #383, opened 2024-10-13
Use migration instead of schema for installs to ensure compatibility with structure.sql

northeastprince:


rosa:
Hey @northeastprince, we purposely switched from migrations to a new DB schema, and aren't planning to revert this. More info at https://github.com/rails/solid_queue/issues/329#issuecomment-2338666058. 
If you need to run a migration for any reason (like to have Solid Queue tables in the same database as your main app), you can do that on your app, copying the schema into a migration for your app.

Issue #210, opened 2024-04-13
Generalize `silence_polling` to `silence_queries`


namolnad:
Yes, I think `silence_queries` would work 👍🏻

_Originally posted by @rosa in https://github.com/rails/solid_queue/issues/198#issuecomment-2048122748_
            
As mentioned in #198, `silence_polling` only silences the polling queries (which makes sense), and continues to log for all other querying methods. I think many, especially in a development environment, would be interested in a `silence_queries` setting to fully silence heart_beat, semaphore, etc queries and keep their logs clean. We likely should maintain `silence_polling`, at least for now, for backwards compatibility.

namolnad:
@rosa do you know of a global way we can wrap all the querying methods in solid_queue in a similar fashion to how the polling is wrapped currently? Or is it more likely that we'll just need to pick off the methods one-by-one and make sure new functionality gets wrapped similarly? Happy to take a stab at this one once that approach gets nailed down. Thanks!

iJackUA:
Up to this issue, as it really quite distracting in dev mode.

With dozens of jobs / recurring jobs and settings 
```
RAILS_LOG_TO_STDOUT=true
RAILS_LOG_LEVEL=debug
```

It's literally an infinite flood in the terminal with SolidQueue related SQL logs.
While developer is mostly interested in few messages types "Job Enqueued" / "Performing Job" / "Performed Job" / "Error trace if there is an error".

I am also a bit confused truing setting different "logger" for SolidQueue, trying to redirect it's logs and using other log_level. But it is not redirecting all logs, only from here https://github.com/rails/solid_queue/blob/main/lib/solid_queue/log_subscriber.rb

Also there is ActiveJob with it's own logger in the equation. And while trying to set custom logger to `SolidQueue.logger` / 
 `ActiveJob::Base.logger` I still could not synchronize them to show me only "significant logs" like `INFO` level.
 
 One workaround for this is 

`bin/jobs`
```ruby
require_relative "../config/environment"
require "solid_queue/cli"

# Keep using single Rails.logger for everything, but for this instance of Ruby env set log level INFO in dev 
Rails.logger.level = Logger::INFO if Rails.env.local?

SolidQueue::Cli.start(ARGV)
```

It keeps logger setting in one place (basic Rails logger config), log levev for web app itself in dev can be still kept `debug`, but SolidQueue now shows me only more or less important "solid_queue/log_subscriber" events and ActiveJob's "enqueue/performed" logs.

[...]

PR #389, opened 2024-10-18
option for silencing logging for heartbeats

jlvallelonga:
In development, and possible other environments, it's sometimes nice to not see the heartbeat logging. This adds an option to silence the logging for heartbeats. It's very similar to the "silence_polling" option.

Usage:

```
  config.solid_queue.silence_heartbeats = true
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ There are several settings that control how Solid Queue works that you can set a
 - `process_alive_threshold`: how long to wait until a process is considered dead after its last heartbeat—defaults to 5 minutes.
 - `shutdown_timeout`: time the supervisor will wait since it sent the `TERM` signal to its supervised processes before sending a `QUIT` version to them requesting immediate termination—defaults to 5 seconds.
 - `silence_polling`: whether to silence Active Record logs emitted when polling for both workers and dispatchers—defaults to `true`.
+- `silence_heartbeats`: whether to silence Active Record logs emitted for heartbeats for both workers and dispatchers—defaults to `false`.
 - `supervisor_pidfile`: path to a pidfile that the supervisor will create when booting to prevent running more than one supervisor in the same host, or in case you want to use it for a health check. It's `nil` by default.
 - `preserve_finished_jobs`: whether to keep finished jobs in the `solid_queue_jobs` table—defaults to `true`.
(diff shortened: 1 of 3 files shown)


rosa:
Thanks @jlvallelonga! However, I was hoping we could get https://github.com/rails/solid_queue/issues/210 done instead of having a bunch of different options to silence different parts 😅

jlvallelonga:
Yeah I was thinking about that a bit when I was doing this. Makes sense :)

madhums:
For me, `silence_polling` and `silence_heartbeat_logging` address different concerns. In that sense, we can keep them as two separate settings. Any chance we can merge this? @rosa

jbescoyez:
In the meantime, I've monkeypatched this PR:

```rb
# /config/initializers

module SilenceHeartbeat
  def heartbeat
    ActiveRecord::Base.logger.silence do
      # Clear any previous changes before locking, for example, in case a previous heartbeat
      # failed because of a DB issue (with SQLite depending on configuration, a BusyException
      # is not rare) and we still have the unpersisted value
      restore_attributes
      with_lock { touch(:last_heartbeat_at) }
    end
  end
end

Rails.application.config.to_prepare do
  SolidQueue::Process.prepend(SilenceHeartbeat) unless Rails.env.production?
end
```

And it woks like a charm! Many tnx @jlvallelonga

mrj:
That work's great Jean-Baptiste to silence the heartbeat logging that happens every minute.

You can also significantly reduce the 5-minute "Prune dead processes"  logs by adding to `config/environments/development.rb` something like
```
config.solid_queue.process_alive_threshold = 1.day
```

plus significantly extend the 10-minute "Unblock jobs" logging by setting in `config/queue.yml`
```
development:
  dispatchers:
    - concurrency_maintenance_interval: 999999
```

which I don't think will cause problems in a non-concurrent development server.

yelvert:
Whats holding up this merge?

rosa:
@yelvert, basically https://github.com/rails/solid_queue/pull/389#issuecomment-2474030096

rosa:
Hey everyone! In the end, I got a related change merged into a Rails to silence all those queries in https://github.com/rails/rails/pull/56899. You just need to set `sql_notifications: false` in the database configuration for Solid Queue. I think this is a better option to silence everything, so I'm going to close this one and document this in the README.

Issue #238, opened 2024-06-05
Default Rails logger does not output to STDOUT in development env


morgoth:
When generating new Rails app (tested with main branch) the deafult Rails logger in development env looks like this:
```
dupa(dev)> Rails.logger
=> 
#<ActiveSupport::BroadcastLogger:0x00007cad6f8fa608
 @broadcasts=
  [#<ActiveSupport::Logger:0x00007cad6f90da78
    @default_formatter=#<Logger::Formatter:0x00007cad6f8fc9a8 @datetime_format=nil>,
    @formatter=#<ActiveSupport::Logger::SimpleFormatter:0x00007cad6f8fa9c8 @datetime_format=nil>,
    @level=0,
    @level_override={},
    @local_level_key=:logger_thread_safe_level_14520,
    @logdev=
     #<Logger::LogDevice:0x00007cad6f90e748
      @binmode=false,
      @dev=#<File:/home/wojtek/Projects/dupa/log/development.log>,
      @filename="/home/wojtek/Projects/dupa/log/development.log",
      @mon_data=#<Monitor:0x00007cad6f8fc8b8>,
      @mon_data_owner_object_id=4560,
      @shift_age=1,
      @shift_period_suffix="%Y%m%d",
      @shift_size=104857600>,
    @progname=nil>,
   #<ActiveSupport::Logger:0x00007cad6f8506a8
    @default_formatter=#<Logger::Formatter:0x00007cad6f1e4800 @datetime_format=nil>,
    @formatter=#<ActiveSupport::Logger::SimpleFormatter:0x00007cad6f1e4698 @datetime_format=nil>,
    @level=0,
    @level_override={},
    @logdev=
     #<Logger::LogDevice:0x00007cad6f850658
      @binmode=false,
      @dev=#<IO:<STDERR>>,
      @filename=nil,
      @mon_data=#<Monitor:0x00007cad6f1e4760>,
      @mon_data_owner_object_id=14500,
      @shift_age=nil,
      @shift_period_suffix=nil,
      @shift_size=nil>,
    @progname=nil>],
 @formatter=#<ActiveSupport::Logger::SimpleFormatter:0x00007cad6f8fa9c8 @datetime_format=nil>,
 @progname="Broadcast">
```

When running SQ via `bundle exec rake solid_queue:start` there is nothing printed on the terminal.
SQ logs are only put into the `log/development.log` file.

I can workaround the issue by specifying logger by hand in `environments/development.rb`:
[...]

PR #390, opened 2024-10-18
Add explicit development instructions to the README.

evdevdev:
I'm playing around with a Rails 8 app, and I had a small bit of confusion in getting Solid Queue running in development. Therefore, I wrote out instructions that I think I would have found helpful. Maybe other folks will find these helpful as well? 

I know that https://github.com/rails/solid_queue/issues/238 mentioned a coming revamp to how Solid Queue works in development. If this clashes with that, I'd be happy to update it to reflect any other changes.


Juksefantomet:
I was unable to get solid queue to work in development until i saw this PR. the Puma.rb edit is vital to get any output regarding development setup. The base note pointing to a plugin mentions nothing about it being required for a native puma+sqlite setup to even catch solid gems.

rosa:
Thanks @evdevdev, sorry for the delay looking into this. I'm still a bit undecided about Solid Queue in development, which is why I haven't merged or requested changes yet. 

@Juksefantomet: 
> The base note pointing to a plugin mentions nothing about it being required for a native puma+sqlite setup to even catch solid gems.

🤔  What do you mean by "required"? Using the Puma plugin is not mandatory, you can run everything in development without it 😕

Juksefantomet:
@rosa 

Prior to adding the plugin entry to puma.rb solid queue config and databases was not created.

Added entries to queue.yml, recurring,yml, database.yml and environment/development.rb did not trigger any results when invoking rails db:prepend or rails db:migrate.

The second i added the plugin entry to puma.rb solid_queue was enabled and database for queue,cable and cache was created for development.

Ubuntu 24.04, rails 8.0.0.rc2, ruby 3.3.3 through rbenv.

Fresh project initialized today as we were prepping a PoC for a new concept at work.

Juksefantomet:
@rosa 

I stand corrected.

The culprit for database generation was not the plugin, but the solid queue is not being used until i add the plugin entry.

The reason for databases not being created was database.yml:

failed config:

```
development:
  <<: *default
  database: storage/development.sqlite3
  cache:
    <<: *default
    database: storage/development_cache.sqlite3
    migrations_paths: db/cache_migrate
  queue:
    <<: *default
    database: storage/development_queue.sqlite3
    migrations_paths: db/queue_migrate
  cable:
    <<: *default
    database: storage/development_cable.sqlite3
    migrations_paths: db/cable_migrate
```

As i simply added the required entries to the existing setup for development i did not add "primary:"

resulting working entry:

```
development:
  primary:
    <<: *default
    database: storage/development.sqlite3
  cache:
    <<: *default
    database: storage/development_cache.sqlite3
    migrations_paths: db/cache_migrate
  queue:
    <<: *default
    database: storage/development_queue.sqlite3
    migrations_paths: db/queue_migrate
  cable:
    <<: *default
    database: storage/development_cable.sqlite3
    migrations_paths: db/cable_migrate
```

but my rails does not provide any console data regarding solid queue and does not trigger internal scheduling until i add the plugin :solid_queue entry to puma.rb

evdevdev:
@rosa Thanks for getting back to me.

If it's helpful, my primary motivation for having this run in development is so that I can get used to all the ergonomics of debugging, reading the logs, monitoring, and all that good stuff. I find myself feeling cautious or timid around having production behave differently from my local setup. I suppose the big fear is that when something catches fire in prod, I won't be able to exactly reproduce it locally. 

Out of curiosity, when you say that you're a bit undecided, what part is up in the air? I really love Solid Queue. Thank you for building it.

evdevdev:
For anyone who finds this PR, I just ran into a little bit of a gotcha. The changes I've proposed in this PR will break broadcasting locally. (In order to fix it, you need to also shift ActionCable into the database rather than `async`.) 

@rosa If you have any interest, I'll happily update this PR to include those instructions as well.

rosa:
Ohhh, @evdevdev, good catch! I think I might be able to figure out this week if Solid Queue configured by default in development is a go or no go, and if it's not, I'll get this PR merged.

rosa:
Thanks a lot for your patience, @evdevdev! Looks like for now things will stay as they are, with the `async` adapter set by default in development, so this should be ready to merge except for the bit about Action Cable using the `async` adapter, that won't work with these settings. If you want to add these instructions to the PR, that'd be awesome. Otherwise I'll add them!

evdevdev:
@rosa Thanks for the update! I'll make the tweak about the Action Cable and add it to this PR before the end of the week.

evdevdev:
@rosa I've updated the README with notes about Action Cable. How's it look to you? Happy to make any tweaks or improvements.

rosa:
This is great, @evdevdev! Thank you so much for doing this! ❤️ 🙏

PR #415, opened 2024-11-17
Add helper method to query the latency on the Queue object

andersonkrs:
Hey all!

I'm using Yabeda to collect some metrics on a rails app that uses solid_queue. I send some metrics like the latency on each queue. Currently I use a little patch to accomplish that. I wonder if we can upstream this path just to make it more convenient for other people to access solid_queue's API. 🤔 

My Yabeda config example

```rb
Yabeda.configure do
  group :solid_queue do
    gauge :queue_latency, unit: :seconds, tags: [:queue]
  end

  collect do
    SolidQueue::Queue.all.each do |queue|
      solid_queue.queue_latency.set({ queue: queue.name }, queue.latency)
    end
  end
end
```

### DIFF
--- a/app/models/solid_queue/queue.rb
+++ b/app/models/solid_queue/queue.rb
@@ -41,7 +41,7 @@ module SolidQueue
     end
 
     def latency
-      @latency = begin
+      @latency ||= begin
         now = Time.current
         oldest_enqueued_at = ReadyExecution.queued_as(name).minimum(:created_at) || now
 
--- a/app/models/solid_queue/queue.rb
+++ b/app/models/solid_queue/queue.rb
@@ -43,7 +43,7 @@ module SolidQueue
     def latency
       @latency = begin
         now = Time.current
-        oldest_enqueued_at = SolidQueue::ReadyExecution.queued_as(name).minimum(:created_at) || now
+        oldest_enqueued_at = ReadyExecution.queued_as(name).minimum(:created_at) || now
 
         (now - oldest_enqueued_at).to_i
       end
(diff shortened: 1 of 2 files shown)


app/models/solid_queue/queue.rb
```diff
@@ -40,6 +40,19 @@ def size
       @size ||= ReadyExecution.queued_as(name).count
     end
 
+    def latency
+      @latency = begin
+        now = Time.current
+        oldest_enqueued_at = SolidQueue::ReadyExecution.queued_as(name).minimum(:created_at) || now
```
rosa, line 46:
```suggestion
        oldest_enqueued_at = ReadyExecution.queued_as(name).minimum(:created_at) || now
```
Just for consistency with the other methods.

app/models/solid_queue/queue.rb
```diff
@@ -40,6 +40,19 @@ def size
       @size ||= ReadyExecution.queued_as(name).count
     end
 
+    def latency
+      @latency = begin
```
rosa, line 44: Was this perhaps intended to be
```ruby
@latency ||= begin
```

rosa (commented):
Hey @andersonkrs, thanks so much for this and sorry for the delay! I just left a couple of minor comments but this is great 🙏

app/models/solid_queue/queue.rb
andersonkrs: ooh good catch! 🙈

andersonkrs: I just fixed it and created a test for the memoization!

andersonkrs:
Hey @rosa  thank you for the review, I just updated the PR with the suggestions!

rosa (approved):
Awesome, thank you so much! 🙏

PR #417, opened 2024-11-18
Reimplement Interruptible

hms:
Details for each of the commits are in the separate commit messages

- Some minor housekeeping with .gitignore, .rubocop.ymlt and solid_queue.gemspec
- Proposed update to the tests to silence successful tests from outputting backtraces (keep the test output clean where possible and only output for the unexpected 
- Proposed reimplementation of the Interruptible concern to help mitigate some mysql database connectivity issues.

### DIFF
--- a/lib/solid_queue/processes/interruptible.rb
+++ b/lib/solid_queue/processes/interruptible.rb
@@ -7,31 +7,25 @@ module SolidQueue::Processes
     end
 
     private
-      SELF_PIPE_BLOCK_SIZE = 11
 
       def interrupt
-        self_pipe[:writer].write_nonblock(".")
-      rescue Errno::EAGAIN, Errno::EINTR
-        # Ignore writes that would block and retry
-        # if another signal arrived while writing
-        retry
+        queue << true
       end
 
       def interruptible_sleep(time)
-        if time > 0 && self_pipe[:reader].wait_readable(time)
-          loop { self_pipe[:reader].read_nonblock(SELF_PIPE_BLOCK_SIZE) }
-        end
-      rescue Errno::EAGAIN, Errno::EINTR
+        # Since this is invoked on the main thread, using some form of Async
+        # avoids a 35% slowdown (at least when running the test suite).
+        #
+        # Using Futures for architectural consistency with all the other Async in SolidQueue.
+        Concurrent::Promises.future(time) do |timeout|
+          if timeout > 0 && queue.pop(timeout:)
+            queue.clear # exiting the poll wait guarantees testing for SHUTDOWN before next poll
+          end
+        end.value
       end
 
-      # Self-pipe for signal-handling (http://cr.yp.to/docs/selfpipe.html)
-      def self_pipe
-        @self_pipe ||= create_self_pipe
-      end
-
-      def create_self_pipe
-        reader, writer = IO.pipe
-        { reader: reader, writer: writer }
+      def queue
+        @queue ||= Queue.new
       end
   end
 end
--- a/lib/solid_queue/processes/interruptible.rb
+++ b/lib/solid_queue/processes/interruptible.rb
@@ -13,13 +13,11 @@ module SolidQueue::Processes
       end
 
       def interruptible_sleep(time)
-        # Since this is invoked on the main thread, using some form of Async
-        # avoids a 35% slowdown (at least when running the test suite).
-        #
-        # Using Futures for architectural consistency with all the other Async in SolidQueue.
+        # Invoking from the main thread can result in a 35% slowdown (at least when running the test suite).
+        # Using some form of Async (Futures) addresses this performance issue.
         Concurrent::Promises.future(time) do |timeout|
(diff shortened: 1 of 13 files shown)


rosa (commented):
Thanks a lot for this PR! I need a little bit of time to review it and I'm on-call this week, so I hope to get to it next week 🙏

rosa:
Thanks again, @hms! Would you mind elaborating a bit more on:
> Proposed reimplementation of the Interruptible concern to help mitigate some mysql database connectivity issues.

What where the connectivity issues you found? Was this only in tests, or did you observe these in a production environment?

hms:
@rosa 

Hi Rosa,

I don't run MySQL in production.  The issues I've come across are strictly in a testing environment.  What I get are random disconnections (and miles long backtraces) at the driver level that are frequently paired with broken pipe errors.  Extra rescues on the pipes side didn't help.  It's always more database errors than pipe errors, leading me to assume the root cause was the database driver.

Since you folks are MySQL based and haven't seen any of these issues, I'm sure this is some weird combination of Mac, my local builds via brew, and/or how I'm configured to build.  I did some digging, and the mysql2 gem had a window of time with similar issues some years ago right after an undocumented MySQL upgrade on the Oracle side, so maybe I'm a cannery in the coal mine?

Relative to a justification for the change, I do think using a pure Ruby solution as long as it's 100% in compliance with the requirements is worth considering.  My understanding for why the self-pipe is to avoid signal processing races that can be triggered by the anything using one of the signal aggregation APIs (database drivers) and signal traps.  The Thread::Queue makes those guarantees with the same performance profile and a very simply to understand implementation.

hms

rosa:
> My understanding for why the self-pipe is to avoid signal processing races that can be triggered by the anything using one of the signal aggregation APIs (database drivers) and signal traps.

Yes, exactly! Thanks for the extra justification; that's super helpful. I do like this implementation much more, so I'm going to run this in production for a couple of days and see if we run into anything, and then I'll merge afterwards. Thank you!

rosa:
Oops, would you mind rebasing against the last `main` and resolve the conflicts? Sorry!

hms:
No problem.  But I won't be able to do for a couple of hours. Just about to step out.

hms

hms:
@rosa

Rebased.

rosa:
This is working great so far, so I'm going to merge it. Thanks a lot @hms!

rosa:
Oops, so sorry, conflicts again! It needs rebasing again 🙈  Also, I realised I missed a comment for a small change, that I'll add now.

lib/solid_queue/processes/interruptible.rb
```diff
@@ -7,31 +7,25 @@ def wake_up
...
 
       def interruptible_sleep(time)
-        if time > 0 && self_pipe[:reader].wait_readable(time)
-          loop { self_pipe[:reader].read_nonblock(SELF_PIPE_BLOCK_SIZE) }
-        end
-      rescue Errno::EAGAIN, Errno::EINTR
+        # Since this is invoked on the main thread, using some form of Async
+        # avoids a 35% slowdown (at least when running the test suite).
+        #
+        # Using Futures for architectural consistency with all the other Async in SolidQueue.
```
rosa, line 19: I think you can remove this last comment as it'd make sense more when evaluating the change, but not after the PR has been merged: 
```suggestion
        # Since this is invoked on the main thread, using some form of Async
        # avoids a 35% slowdown (at least when running the test suite).
```

hms: @rosa 

I know the SQ team isn't in the code comment camp.  Most, if not all of my comments, were there to help the PR review.  I'm happy to update or remove as preferred.

solid_queue.gemspec
```diff
@@ -44,4 +44,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "pg"
   spec.add_development_dependency "sqlite3"
   spec.add_development_dependency "rubocop-rails-omakase"
+  spec.add_development_dependency "ostruct"
+  spec.add_development_dependency "logger"
+  spec.add_development_dependency "fiddle"
```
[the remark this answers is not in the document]
hms: @rosa 

Because without it, running rails console was triggering warnings.  However, those warnings look to be tied to a Gem version that has subsequently updated between start and finish of this PR.  I'll make a pass on those 3 additions and verify if they are still required or not.

hms:
@rosa 

Rebased. 

Thank you for all of your feedback and help to get this finished.

lib/solid_queue/processes/interruptible.rb
rosa: Thank you so much! The comments to help with the PR review are super appreciated! If you want, next time you can also leave the comments using GitHub comments after you've submitted the PR. This is something I do myself, so I don't have to commit the comments to the code but still help with the PR review.

rosa:
Thank you!

Issue #421, opened 2024-11-20
Jobs and transactional integrity docs don't seem to be what actually happens in Solid Queue when using a single DB


leondmello:
> Having your jobs in the same ACID-compliant database as your application data enables a powerful yet sharp tool: taking advantage of transactional integrity to ensure some action in your app is not committed unless your job is also committed and viceversa, and ensuring that your job won't be enqueued until the transaction within which you're enqueuing it is committed.

The above text seems to indicate that if you use the same DB for `ActiveRecord` application models and the Solid Queue models, job enqueuing will only happen if the application model changes are successfully committed to the DB & vice versa.
We changed our implementation to move from a separate DB to a single DB to avail of this. But in practice, we are seeing that this is not the case.

After we installed some debuggers, we found that eventually, [this line of code is hit](https://github.com/rails/solid_queue/blob/393e3b9eff7210f42a43d7e6456b7a172ba78d42/app/models/solid_queue/job/executable.rb#L104). 
[That rails method](https://github.com/rails/rails/blob/7750d64a65e5b2641d87ef45e6e65ace193d9a27/activerecord/lib/active_record/relation.rb#L283) creates a new nested transaction.

Such nested transactions will not roll back all transactions correctly [when using something like postgres](https://github.com/rails/rails/blob/7750d64a65e5b2641d87ef45e6e65ace193d9a27/activerecord/lib/active_record/transactions.rb#L135-L174). We confirmed this behaviour by explicitly rolling back the application transaction after enqueuing the job. The Job record was saved, the application model record was not.

leondmello:
An extension to this issue: https://github.com/rails/solid_queue/issues/409

rosa:
Hey @leondmello, sorry for the delay, I'm on-call this week. 

> But in practice, we are seeing that this is not the case.

Do you have a minimal reproduction for this where you're seeing this is not the case due to the nested transactions? 

> Maybe a simple fix is to split that up into a `find` & `create!`

Unfortunately, that's not the same as `create_or_find_by!`, as the `find` might return a stale read and then the `create!` fails with a duplicate key exception.

rosa:
> An extension to this issue: https://github.com/rails/solid_queue/issues/409

[...]

PR #424, opened 2024-11-21
Do not create a new transaction when creating a ReadyExecution

leondmello:
Fixes https://github.com/rails/solid_queue/issues/421

### DIFF
--- a/app/models/solid_queue/job/executable.rb
+++ b/app/models/solid_queue/job/executable.rb
@@ -101,7 +101,9 @@ module SolidQueue
 
       private
         def ready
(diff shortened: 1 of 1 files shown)


leondmello:
@rosa Does this line need the same change?
https://github.com/rails/solid_queue/blob/393e3b9eff7210f42a43d7e6456b7a172ba78d42/app/models/solid_queue/job/schedulable.rb#L40

rosa:
Closing per https://github.com/rails/solid_queue/issues/421#issuecomment-2494928082.

PR #425, opened 2024-11-24
Implements TimerTasks using Promises.  

hms:
Replaces the OG TimerTask with a Promises based implementation.

In combination with the previous commit, all of SQ has been moved
to the Promises infrastructure and makes using any of the new 
concurrency functionality (Actors, Channels, etc.) more easily available 
for future SQ enhancements.

This PR is a continuation of the not yet accepted PR #417.  If desired / required,
I am happy to pull the 2 commits into a clean PR based on main.

### DIFF
--- a/app/models/solid_queue/claimed_execution.rb
+++ b/app/models/solid_queue/claimed_execution.rb
@@ -5,12 +5,6 @@ class SolidQueue::ClaimedExecution < SolidQueue::Execution
 
   scope :orphaned, -> { where.missing(:process) }
 
-  class Result < Struct.new(:success, :error)
-    def success?
-      success
-    end
-  end
-
   class << self
     def claiming(job_ids, process_id, &block)
       job_data = Array(job_ids).collect { |job_id| { job_id: job_id, process_id: process_id } }
@@ -60,12 +54,9 @@ class SolidQueue::ClaimedExecution < SolidQueue::Execution
   def perform
     result = execute
 
-    if result.success?
-      finished
-    else
-      failed_with(result.error)
-      raise result.error
-    end
+    result.just? ? finished : failed_with(result.reason)
+
+    result
   ensure
     job.unblock_next_blocked_job
   end
@@ -93,9 +84,9 @@ class SolidQueue::ClaimedExecution < SolidQueue::Execution
   private
     def execute
       ActiveJob::Base.execute(job.arguments)
-      Result.new(true, nil)
+      Concurrent::Maybe.just(true)
     rescue Exception => e
-      Result.new(false, e)
+      Concurrent::Maybe.nothing(e)
     end
 
     def finished
--- a/test/models/solid_queue/claimed_execution_test.rb
+++ b/test/models/solid_queue/claimed_execution_test.rb
@@ -22,9 +22,7 @@ class SolidQueue::ClaimedExecutionTest < ActiveSupport::TestCase
     job = claimed_execution.job
 
     assert_difference -> { SolidQueue::ClaimedExecution.count } => -1, -> { SolidQueue::FailedExecution.count } => 1 do
-      assert_raises RuntimeError do
-        claimed_execution.perform
-      end
+      claimed_execution.perform
     end
 
     assert_not job.reload.finished?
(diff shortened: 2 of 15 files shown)


rosa:
Hi @hms, so sorry for the delay here! I didn't have the chance to look at this and #417 this past week as I focused on something else, but this is top of my list for next week.

hms:
@rosa 

Rebased this PR too.

rosa:
Hey @hms, would you mind rebasing this one once more, now that #417 has been merged?

app/models/solid_queue/claimed_execution.rb
```diff
@@ -60,12 +54,9 @@ def discard_all_from_jobs(*)
...
 
-    if result.success?
-      finished
-    else
-      failed_with(result.error)
-      raise result.error
-    end
+    result.just? ? finished : failed_with(result.reason)
+
+    result
```
rosa, line 59: What's the advantage of using `Concurrent::Maybe` here vs. a regular struct?

hms: Warning: Rationalization alert.

I dislike Struct.  I would have reached for Data, but when I started this PR, SQ was on Ruby 3.3.1.

The Result struct was nothing more than a simple Maybe.  It was also only used in 1 place, for 1 `if` and made me question if it was really needed at all (return Array[success, exception]).  But with the change from Futures to Promises::Futures, I also pushed the error reporting out of the execute method back to the Future. With some of the logic moving back to the Future, why keep the logic all within the Concurrent library?

If all of this is bothersome, you're not buying the rationalization for the change, of you simply feel that this doesn't make SQ better I'm happy to refactor the Maybe out and, now that SQ is using Ruby 3.3.5, bring the Result object back, but as a Data instead of Struct.

hms:
rebased

hms:
@rosa 

One other question should this PR be accepted.  I avoided using Concurrent::Cancellation because it's in Edge.  But it's been in edge since 2018, is the SQ team open my cleaning up the TimerTask implementation using Cancellation?  If yes, the code would be a little cleaner and it would allow the last bit of the older concurrency implementation used in Scheduler to be reimplemented via the Promises infrastructure.

test/models/solid_queue/claimed_execution_test.rb
```diff
@@ -22,9 +22,7 @@ class SolidQueue::ClaimedExecutionTest < ActiveSupport::TestCase
     job = claimed_execution.job
 
     assert_difference -> { SolidQueue::ClaimedExecution.count } => -1, -> { SolidQueue::FailedExecution.count } => 1 do
-      assert_raises RuntimeError do
-        claimed_execution.perform
-      end
+      claimed_execution.perform
```
rosa, line 25: This was added in https://github.com/rails/solid_queue/pull/373, and it's not clear to me that the new behaviour is correct 🤔  If the error reporting library doesn't rely on Rails error reporting, I think this might not work correctly.

hms: @rosa 

I'm not sure I fully understand your issue / question or possibly the impact of #373.  

Is your issue with the actual error handling and possibly missed / lost errors, this test is not valid with my changes and hence I'm no longer testing an important behavior, or what happens if a SQ user is depending on something other than Rails.error for their system wide Error manage mechanism?

hms

hms: @rosa 

I've done some digging into ActiveSupport::ExecutorWrapper and SolidQueue initialization and I think I understand your issue/concern. 

When an exception is raised and not rescued within a ExecutionWrapper.wrap block, it's caught and reported to the ActiveSupport.error_reporter.  Under normal circumstances, ActiveSupport.error_reporter and Rails.error_reporter are one and the same.  So, the code in the PR is fine.... under normal circumstances.

However, if directly or indirectly, the app_executor has been overridden and the overridden app_executor handles exception reporting without calling Rails.error_reporter, then the current implementation of the PR is effectively swallowing the exception.

If this summary of your concern is correct, I'll see what I have to do to address it.

hms:
@rosa 

Based on your feedback and a clear misunderstanding of a Concurrent Ruby behavior, I'm going to pull this PR and re-submit it.  I'll put back the Results, since with the required rework, Maybe is even harder to justify.   Thanks for your patience and catching the significant issue with the code as submitted.

PR #426, opened 2024-11-26
Upgrade mysql2 gem

jherdman:
I ran into the issues described here when trying to install necessary
gems to contribute:

https://github.com/brianmario/mysql2/issues/1350

Upgrading seems to have resolved the issue.


rosa:
Thanks @jherdman!

PR #427, opened 2024-11-26
Error When an Invalid Recurring Task is Configured

### LINKED ISSUE #414: Invalid Recurring Jobs: No Feedback
I recently added a new entry to `recurring.yml` and mistyped the name of the corresponding job class. I only realized something was askew when I viewed the Recurring Tasks tab in Mission Control.

I believe there ought to be some feedback when an invalid job class is listed. Perhaps an error or warning. This is something I'm happy to try and implement if it's of interest.

jherdman:
I took a first crack at resolving #414. I think it's a little bit of a blunt solution.

Resolves #414

### DIFF
--- a/lib/solid_queue/configuration.rb
+++ b/lib/solid_queue/configuration.rb
@@ -36,6 +36,14 @@ module SolidQueue
       end
     end
 
+    def valid?
+      skip_recurring_tasks? || recurring_tasks.none? || recurring_tasks.all?(&:valid?)
+    end
+
+    def invalid_tasks
+      recurring_tasks.select(&:invalid?)
+    end
+
     def max_number_of_threads
       # At most "threads" in each worker + 1 thread for the worker + 1 thread for the heartbeat task
       workers_options.map { |options| options[:threads] }.max + 2
@@ -100,7 +108,7 @@ module SolidQueue
       def recurring_tasks
         @recurring_tasks ||= recurring_tasks_config.map do |id, options|
           RecurringTask.from_configuration(id, **options)
-        end.select(&:valid?)
+        end
       end
 
       def processes_config
--- a/lib/solid_queue/configuration.rb
+++ b/lib/solid_queue/configuration.rb
@@ -37,11 +37,21 @@ module SolidQueue
     end
 
     def valid?
-      skip_recurring_tasks? || recurring_tasks.none? || recurring_tasks.all?(&:valid?)
+      configured_processes.any? && (skip_recurring_tasks? || invalid_tasks.none?)
     end
 
-    def invalid_tasks
-      recurring_tasks.select(&:invalid?)
+    def error_messages
+      if configured_processes.none?
+        "No workers or processed configured. Exiting..."
+      else
+        error_messages = invalid_tasks.map do |task|
+            all_messages = task.errors.full_messages.map { |msg| "\t#{msg}" }.join("\n")
+            "#{task.key}:\n#{all_messages}"
+          end
+          .join("\n")
+
+        "Invalid processes configured:\n#{error_messages}"
+      end
     end
 
     def max_number_of_threads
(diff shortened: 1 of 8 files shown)


lib/solid_queue/scheduler.rb
```diff
@@ -35,7 +35,15 @@ def run
       end
 
       def schedule_recurring_tasks
-        recurring_schedule.schedule_tasks
+        if recurring_schedule.valid?
+          recurring_schedule.schedule_tasks
+        else
+          error_messages = recurring_schedule.invalid_tasks.map do |task|
```
jherdman, line 41: It occurs to me that if you have a record written in your DB, and then it _becomes_ invalid this will error out. E.g. your record has a legacy class name. This would prevent SQ from booting, which might be undesirable. Perhaps it's best to log the errors instead?

rosa:
Thanks a lot @jherdman! I was thinking that perhaps this should be part of a more generic configuration validation that runs and raises before starting the supervisor 🤔

jherdman:
Perhaps we need both things? Since the recurring jobs are just DB backed, and have validations, we probably need to report on those validation errors should they occur.

rosa:
> we probably need to report on those validation errors should they occur.

Right, I mean doing this as part of that configuration validation that runs before the supervisor starts, instead of waiting until we try to schedule them.

rosa:
@jherdman thank you! I was just thinking of having the current check [here](https://github.com/rails/solid_queue/blob/be8f367b0603496f2fd9b2a7bc97ca1e8866cbfd/lib/solid_queue/supervisor.rb#L11-L17) expanded to also check if the configuration is valid (`configuration.valid?`), and if not, abort printing the existing errors. I think not having any configured processes could be part of the `configuration.valid?` check.

lib/solid_queue/configuration.rb
```diff
@@ -36,6 +36,14 @@ def configured_processes
       end
     end
 
+    def valid?
+      skip_recurring_tasks? || recurring_tasks.none? || recurring_tasks.all?(&:valid?)
```
jherdman, line 40: I feel like this is probably not exhaustive enough, but I'm unsure of other cases to expand upon. This is probably sufficient for the scope of this PR though.

test/unit/supervisor_test.rb
```diff
@@ -48,6 +48,14 @@ class SupervisorTest < ActiveSupport::TestCase
     assert_not process_exists?(pid)
   end
 
+  test "start with invalid configuration" do
```
jherdman, line 51: I wasn't entirely sure how to test the abort message. Perhaps that's OK though? There are no other tests for the specific abort message.

jherdman (commented):
@rosa take two! I think this is more in line with what you noted above.

test/unit/supervisor_test.rb
jherdman: Ah! A way forward on making this test a little better: https://docs.seattlerb.org/minitest/Minitest/Assertions.html#method-i-capture_subprocess_io

lib/solid_queue/configuration.rb
rosa: Could we include perhaps the case of no configured processes? So we could unify that call in the Supervisor, instead of having two separate checks? 

Besides that, I think you can skip `recurring_tasks.none?` because if `recurring_tasks` is empty, `recurring_tasks.all?(&:valid?)` will return `true`.

jherdman: Ugh... seems so obvious after you pointed it out 😅 

I hope you don't mind, but I've kept the two distinct error messages. It feels like the verbose invalid task error message would be helpful.

rosa (approved):
Wonderful! Thank you so much, @jherdman! 🙏 🙇‍♀️

Issue #420, opened 2024-11-20
Recurring Tasks - still confused


Nilpointer:
Hi,

I am still unsure how to manage reccuring tasks. 

- How do I delete them? ( I don't understand the provided solutions - sorry)
- How can I pause them? Can I?
- How can I update the schedule value at runtime with no downtime?

Can I simply edit the solid_queue_recurring_tasks schedule value or does this cause unwanted sideeffects?

Regards 
Andreas

rosa:
> How do I delete them? ( I don't understand the provided solutions - sorry)

You delete them from your `recurring.yml` configuration (or another file if you are configuring them elsewhere). 

> How can I pause them? Can I?

No, pausing them is not supported. You'd need to remove them and then re-add them. 

> How can I update the schedule value at runtime with no downtime?

You'd need to have a no-downtime deploy as you need to update your configuration and deploy your code. 

> Can I simply edit the solid_queue_recurring_tasks schedule value or does this cause unwanted sideeffects?

This won't work in general because in most cases the scheduler will have the tasks already loaded from when it started, and then when it restarts, it'll read the configuration and would load any task there again, overwriting your changes in `solid_queue_recurring_tasks`.

rosa:
> But how? If I remove the development node from my schedule.yml, restart the application - the entries in the solid_queue_recurring_tasks table are still there.

Yes, sorry, they're there but it doesn't mean they're running. See https://github.com/rails/solid_queue/issues/382. I'll change that so they get deleted as well when the scheduler reloads again. I just haven't had time this week.

Nilpointer:
Is there a way to inspect/query what jobs are currently scheduled?

rosa:
@Nilpointer just checking the contents of `recurring.yml`. I'm going to fix this shortly, in any case.

rosa:
This one should be now handled by #433.

PR #433, opened 2024-11-29
Delete static recurring tasks before loading them again

### LINKED ISSUE #382: Should delete the old recurring task when the task changed key, or removed
When the key of a recurring task is changed or the task is removed, should the old task be automatically deleted? This is to ensure that outdated tasks do not continue to run.

rosa:
Closes #382 and #420.


rosa:
Hey @HLFH, yes! Sorry for the delay, I had been distracted with other work, but I'll ship a new release today.

rosa:
Released v1.1.0 with that change, among others 😊

PR #447, opened 2024-12-09
Specify `source` of SolidQueue to ErrorReporter

### LINKED ISSUE #446: Specify `source` of SolidQueue to ErrorReporter
Because SolidQueue uses the [Rails executor](https://guides.rubyonrails.org/threading_and_code_execution.html#executor), exceptions raised in a Job are automatically handled by [ActiveSupport::ErrorReporter](https://guides.rubyonrails.org/error_reporting.html).

This is very convenient.
I want to make this even more convenient.

## Proposal

Simply put, this is what I propose.
```diff
diff --git a/lib/solid_queue/app_executor.rb b/lib/solid_queue/app_executor.rb
index da0976f..0580213 100644
--- a/lib/solid_queue/app_executor.rb
+++ b/lib/solid_queue/app_executor.rb
@@ -4,7 +4,7 @@ module SolidQueue
   module AppExecutor
     def wrap_in_app_executor(&block)
       if SolidQueue.app_executor
-        SolidQueue.app_executor.wrap(&block)
+        SolidQueue.app_executor.wrap(source: "application.solid_queue", &block)
       else
         yield
       end
```

Currently, the [default value](https://github.com/rails/rails/blob/v8.0.0/activesupport/lib/active_support/execution_wrapper.rb#L86) `application.active_support` is applied for source.
This does not identify it as an exception raised inside SolidQueue.

This is a quote from the [rails guide](https://guides.rubyonrails.org/error_reporting.html#error-reporting-options).
> source: a String about the source of the error. The default source is "application". Errors reported by internal libraries may set other sources; the Redis cache library may use "redis_cache_store.active_support", for instance. Your subscriber can use the source to ignore errors you aren't interested in.

Again, examples of the use of `source` were explained.
https://github.com/rails/rails/pull/44999

What do you think?

tmimura39:
Currently, the default value `application.active_support` is applied for source. This does not identify it as an exception raised inside SolidQueue.

> source: a String about the source of the error. The default source is "application". Errors reported by internal libraries may set other sources; the Redis cache library may use "redis_cache_store.active_support", for instance. Your subscriber can use the source to ignore errors you aren't interested in.

This is a quote from the [rails guide](https://github.com/rails/rails/blob/v8.0.0/guides/source/error_reporting.md?plain=1#L216-L220).

Follow this guide to specify the source so that it can be determined that the error originated from inside SolidQueue.

Fixes #446

### DIFF
--- a/lib/solid_queue/app_executor.rb
+++ b/lib/solid_queue/app_executor.rb
@@ -4,7 +4,7 @@ module SolidQueue
   module AppExecutor
     def wrap_in_app_executor(&block)
       if SolidQueue.app_executor
-        SolidQueue.app_executor.wrap(&block)
+        SolidQueue.app_executor.wrap(source: "application.solid_queue", &block)
       else
         yield
       end


lib/solid_queue/app_executor.rb
```diff
@@ -4,7 +4,7 @@ module SolidQueue
   module AppExecutor
     def wrap_in_app_executor(&block)
       if SolidQueue.app_executor
-        SolidQueue.app_executor.wrap(&block)
+        SolidQueue.app_executor.wrap(source: "application.solid_queue", &block)
```
tmimura39, line 7: > application.solid_queue

Not sure if these values are the best🙃 

I followed the naming rules described [here](https://github.com/rails/rails/pull/44999#discussion_r873140565).

rosa: That seems reasonable to me, reading that comment there 🤔

rosa:
This is great, thank you so much @tmimura39! 🙇‍♀️ 🙏

PR #452, opened 2024-12-11
Add a note about `threads`

genya0407:
I wrote a note explaining the relationship between the database connection pool size and the `threads` parameter to guide users on tuning the `threads` parameter.

For more details, see the related discussion: https://github.com/rails/solid_queue/pull/449


epugh:
One thing that has been a learning for me is that with the various solid gems, I need more resources for the database.  The app I steward is deployed on Heroku with a fairly small mysql database from a managed service vendor...   I've had to go up in the size of the database in order to get more support for number of connections.   I do *love* the simplicity of not needing another system like redis to learn, but did need to increase my mysql resources ;-).

rosa (approved):
Thank you! 🙏

PR #457, opened 2024-12-17
Validate DB connection pool before starting supervisor

rosa:
Small follow-up to #427 to also fail when the DB connection pool is too small.

### DIFF
--- a/lib/solid_queue/configuration.rb
+++ b/lib/solid_queue/configuration.rb
@@ -2,6 +2,12 @@
 
 module SolidQueue
   class Configuration
+    include ActiveModel::Model
+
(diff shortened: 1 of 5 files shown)


jherdman:
Neat idea to use Active Model

ehutzelman:
Just wanted to drop a note in case others run into the same issue I did with this change. For those running Kamal deployments and using the RAILS_MAX_THREADS as the pool size, the job runner process will fail to start but Kamal thinks it's a good deployment because the web process starts fine. The jobs docker container will be stuck in an infinite restart loop until the pool size is appropriately configured.

PR #486, opened 2025-01-16
Add lifecycle hooks to Dispatcher and Scheduler

hms:
I'm switching over to Rails Semantic Logger and for that to work with forking code it requires a post fork action.

With this change, all 4 process types have start and stop hooks and it becomes possible to do on_*_start { post_fork_action } and on_*_stop { pre_stop_action }


rosa:
Hey @hms, sorry for the delay on this one! Nice!

PR #496, opened 2025-01-24
Fix issue 482

hms:
- Addresses Ruby version issues.
  - Ruby 3.1.6 Thread::Queue.pop does not support a timeout option.  To address this issue, given that Ruby 3.1.6 is 2 months from EOL, this PR conditionally includes either the current version of Interruptible or the original implementation of Interruptible.  It's intended this part of the PR should be removed once SQ raising the min supported Ruby to >= 3.2
  - Fixes claimed_execution_test to support Ruby 3.4+ (different backtrace formatting)
  - Updates the Github actions to test a major subset of the rubies between 3.1.6 - 3.4.1.
    - Required special case handling for Ruby 3.1.6 because of a dependency issue with Zeitwerk.
 - Added error reporting in Interruptible#interruptible_sleep.  In theory, this shouldn't ever be needed, but since I thought that with the original PR....
 - Addresses a silly little bug in ScheduledExecution that didn't have an effect, but it was a bug.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 name: Build
-on: [push, pull_request]
+on: [ push, pull_request ]
 
 jobs:
   rubocop:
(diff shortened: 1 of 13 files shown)


rosa:
Thanks a lot @hms! 🙏

PR #503, opened 2025-01-29
feat: add on_exit hook

ThomasCrambert:
Expose new `on_exit` hook to allow post-shutdown cleanup actions.

Notes:
- I made it for all kind of processes, but I believe it would only be useful for workers as they are the only one that may require such hooks. Do you believe it should only be for workers now and add them for the others later?
- Dispatchers currently trigger `run_stop_hooks` in [after_shutdown](https://github.com/rails/solid_queue/blob/cb5d230ddbf6732f2b78a5d3748475bab3b1deb5/lib/solid_queue/dispatcher.rb#L11)  but workers trigger them at [before_shutdown](https://github.com/rails/solid_queue/blob/cb5d230ddbf6732f2b78a5d3748475bab3b1deb5/lib/solid_queue/worker.rb#L8), I changed it in this PR but was that the expected behavior?

### DIFF
--- a/lib/solid_queue.rb
+++ b/lib/solid_queue.rb
@@ -41,30 +41,20 @@ module SolidQueue
   mattr_accessor :clear_finished_jobs_after, default: 1.day
   mattr_accessor :default_concurrency_control_period, default: 3.minutes
 
-  delegate :on_start, :on_stop, to: Supervisor
+  delegate :on_start, :on_stop, :on_exit, to: Supervisor
 
-  def on_worker_start(...)
-    Worker.on_start(...)
-  end
+  [ Dispatcher, Scheduler, Worker ].each do |process|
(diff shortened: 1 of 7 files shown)


rosa:
Thanks @ThomasCrambert!

> Dispatchers currently trigger run_stop_hooks in [after_shutdown](https://github.com/rails/solid_queue/blob/cb5d230ddbf6732f2b78a5d3748475bab3b1deb5/lib/solid_queue/dispatcher.rb#L11) but workers trigger them at [before_shutdown](https://github.com/rails/solid_queue/blob/cb5d230ddbf6732f2b78a5d3748475bab3b1deb5/lib/solid_queue/worker.rb#L8), I changed it in this PR but was that the expected behavior?

Oh! No, that was an oversight. Thanks for catching that 🙏

rosa:
> Do you believe it should only be for workers now and add them for the others later?

No, I think this is fine 👍

PR #505, opened 2025-01-30
Enable multisharding for solid queue

### LINKED ISSUE #353: DB Shards support
Hi,

In https://github.com/rails/solid_queue/pull/56 it seems built in support for database sharding was added to solid_queue. 

In my team we were unsuccessful in testing this feature. Maybe we are not setting things up correctly. What we did we set up in the config `config.solid_queue.connects_to` to point to two shards as in:
```
config.solid_queue.connects_to shards: {
    shard_one: { writing: :primary_shard_one, reading: :primary_shard_one_replica },
    shard_two: { writing: :primary_shard_two, reading: :primary_shard_two_replica }
  }
```

We tried the configuration above, but the workers and the dispatchers only connected to the first shard. We managed to get the adapter enqueue in both shards with some tweaking, but it did not support it out of the box.

If the sharding is supported out of the box, would you be so kind to point us to a example of DB sharding working out of the box (configuration for adapter, dispatcher and worker)? Or, if the sharding is not fully supported yet, is there any plan to enhance solid queue with full sharding support ?

Many thanks!

Regards,
Andrei

alxckn:
Fixes #353

This is a proposal to add multisharding as a supported configuration option for Solid Queue. It makes it possible to easily migrate incrementally an application from one to multiple shards as changing the `connects_to` configuration will not change any of the default behavior of Solid Queue.

This introduces 2 ways to determine which shard to enqueue a job to:
- specify a configured `queue_adapter` on a job, so that a given job class would be assigned to a given shard (ensuring full compatibility with some other Solid Queue features such as concurrency control)
- specify a global `shard_selection_lambda` to have complete control over shard selection logic when enqueueing jobs

By configuring the `active_shard` dynamically (for instance using an environment variable), it is possible to start workers, dispatchers and schedulers that will consume each shard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -618,6 +618,73 @@ my_periodic_resque_job:
 
 and the job will be enqueued via `perform_later` so it'll run in Resque. However, in this case we won't track any `solid_queue_recurring_execution` record for it and there won't be any guarantees that the job is enqueued only once each time.
 
+## Multisharding
+
+If your application reaches a point where the pressure on the database used for jobs is such that you need to spread the load over multiple databases, then this section is for you.
+
+You can extend the Solid Queue database configuration to use different shards:
+
+  ```ruby
+  config.solid_queue.connects_to = {
+    shards: {
+      queue_shard_one: { writing: :queue_shard_one },
+      queue_shard_two: { writing: :queue_shard_two }
+    }
+  }
+  ```
+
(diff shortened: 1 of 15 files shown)


rosa:
Hey @alxckn, thanks a lot for starting this and sorry for the delay. I was supposed to work on sharding at the end of last year but in the end I switched to work on something else and have been focused on other stuff since then. 

I'll look into this soon.

PR #511, opened 2025-02-05
Make sure to close DB connections before fork

### LINKED ISSUE #506: Fork safety warning when booting Solid Queue
In the context of a brand new app with Solid Queue installed (SQlite), a warning is emitted when booting the supervisor using `bin/jobs start`:

```
.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/sqlite3-2.5.0-x86_64-linux-gnu/lib/sqlite3/fork_safety.rb:43: warning: Writable sqlite database connection(s) were inherited from a forked process. This is unsafe and the connectio
ns are being closed to prevent possible data corruption. Please close writable sqlite database connections before forking.
```

The origin could be https://github.com/rails/solid_queue/blob/multisharding/lib/solid_queue/configuration.rb#L79-L79, a connection to the Solid Queue DB is opened when the app is being initialized and isn't be explicitely closed before the supervisor forks workers, dispatchers or schedulers.

alxckn:
Fixes #506. This change helps remove the warning issued by sqlite3 gem

### DIFF
--- a/app/models/solid_queue/record.rb
+++ b/app/models/solid_queue/record.rb
@@ -6,6 +6,11 @@ module SolidQueue
 
     connects_to(**SolidQueue.connects_to) if SolidQueue.connects_to
 
+    def self.clear_all_connections!
+      self.connection_handler.clear_all_connections!(:writing)
+      self.connection_handler.clear_all_connections!(:reading)
+    end
+
     def self.non_blocking_lock
       if SolidQueue.use_skip_locked
         lock(Arel.sql("FOR UPDATE SKIP LOCKED"))
--- a/lib/puma/plugin/solid_queue.rb
+++ b/lib/puma/plugin/solid_queue.rb
@@ -12,7 +12,7 @@ Puma::Plugin.create do
     end
 
     launcher.events.on_booted do
-      @solid_queue_pid = fork do
+      @solid_queue_pid = SolidQueue.safe_fork do
         Thread.new { monitor_puma }
         SolidQueue::Supervisor.start
       end
--- a/lib/puma/plugin/solid_queue.rb
+++ b/lib/puma/plugin/solid_queue.rb
@@ -12,7 +12,7 @@ Puma::Plugin.create do
     end
 
     launcher.events.on_booted do
-      @solid_queue_pid = SolidQueue.safe_fork do
+      @solid_queue_pid = fork do
         Thread.new { monitor_puma }
         SolidQueue::Supervisor.start
       end
--- a/lib/solid_queue.rb
+++ b/lib/solid_queue.rb
@@ -73,5 +73,10 @@ module SolidQueue
     ActiveSupport::Notifications.instrument("#{channel}.solid_queue", **options, &block)
   end
 
+  def safe_fork(&block)
+    Record.clear_all_connections!
+    fork { block.call }
+  end
+
   ActiveSupport.run_load_hooks(:solid_queue, self)
 end
--- a/lib/solid_queue.rb
+++ b/lib/solid_queue.rb
@@ -20,6 +20,8 @@ loader.setup
 module SolidQueue
   extend self
 
+  ForkSafety # add fork safety hooks
+
   DEFAULT_LOGGER = ActiveSupport::Logger.new($stdout)
 
   mattr_accessor :logger, default: DEFAULT_LOGGER
@@ -71,11 +73,6 @@ module SolidQueue
 
   def instrument(channel, **options, &block)
     ActiveSupport::Notifications.instrument("#{channel}.solid_queue", **options, &block)
-  end
-
-  def safe_fork(&block)
-    Record.clear_all_connections!
-    fork { block.call }
   end
 
   ActiveSupport.run_load_hooks(:solid_queue, self)
--- a/lib/solid_queue/fork_safety.rb
+++ b/lib/solid_queue/fork_safety.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module SolidQueue
+  module ForkSafety
+    def _fork
+      Record.clear_all_connections!
+
+      pid = super
+
+      pid
+    end
+  end
+end
+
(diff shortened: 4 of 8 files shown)


app/models/solid_queue/record.rb
```diff
@@ -6,6 +6,11 @@ class Record < ActiveRecord::Base
 
     connects_to(**SolidQueue.connects_to) if SolidQueue.connects_to
 
+    def self.clear_all_connections!
+      self.connection_handler.clear_all_connections!(:writing)
+      self.connection_handler.clear_all_connections!(:reading)
```
flavorjones, line 11: I'm curious why two calls are being made for reading and writing roles here.

By default, if no arguments are passed to `ConnectionHandler#clear_all_connections!` it should clear all roles. Did you see something unexpected that led you to code it this way?

flavorjones (commented):
👋 sqlite3-ruby maintainer here. The general approach is sound, I just have a question about the connection pool roles.

app/models/solid_queue/record.rb
alxckn: Hello, to be honest I was just copy pasting from another app without digging the reason.
The behavior for this method changes in rails 7.2 according to [this PR](https://github.com/rails/rails/commit/64cbcd7a8c0d83f0da48fdd01bd156aaf12c8221#diff-d1e52309089a1c9135a7d1bb82bdfdf7410d7d7a294013e16e71f0171154eefbL196-L199). I just checked an solid queue still supports rails 7.1, so that would ensure compatibility.

rosa:
Thanks @alxckn! Looks like [this breaks the tests for SQLite](https://github.com/rails/solid_queue/actions/runs/13159483494/job/36730103712?pr=511) for some Ruby versions (3.2.0 and 3.3.2, no idea why just these), with some quite odd DB-related failures (I hadn't seen these before). Do you have any idea why this might be the case? I suspect this is some artificial consequence of the tests, but I have no idea what 😕

alxckn:
@rosa , I don't know why the tests fail either, I'll find time to dig

app/models/solid_queue/record.rb
flavorjones: @alxckn Ah! That's a very good reason to do it that way. Thanks for the explanation! 👍

alxckn:
@rosa I extended the disconnect to all forks made in the lib, that seems to have taken care of the initial sqlite failures.

It then made most mysql tests fail, turned out the change affected this [processes_lifecycle_test](https://github.com/rails/solid_queue/pull/511/commits/57036dcebcf930c20cda67f6ce1bf15aec793ca8), the supervisor would [deregister](https://github.com/alxckn/solid_queue/blob/close_connection_before_fork_local/app/models/solid_queue/process.rb#L30-L41) and so [move the job from claimed to ready](https://github.com/alxckn/solid_queue/blob/close_connection_before_fork_local/app/models/solid_queue/claimed_execution.rb#L74-L81) before the test could verify it [remained claimed](https://github.com/rails/solid_queue/pull/511/commits/57036dcebcf930c20cda67f6ce1bf15aec793ca8#diff-640517dba3f5f45f286a944bd047634aaaeb8d83d400a77b5f272f1d7bf3aa1eL138). Adjusting the test to sleep a bit less seems to fix it, but probably in a (more?) flaky way. 

Now there are still failing runs, because of DB deadlocks (ex. [sqlite](https://github.com/rails/solid_queue/actions/runs/13209128855/job/36879081414?pr=511), [PG](https://github.com/alxckn/solid_queue/actions/runs/13209128318/job/36879076713)). Looks scary, they seem to go away upon retry. I wonder if that's something that you've already seen happen within this repo, or if it's definitely something that this PR is bringing?

rosa:
I guess we need to ask: why do we need to do this here? [Rails is supposed to handle this scenario](https://github.com/rails/rails/blob/00cc4ff0259c0185fe08baadaa40e63ea2534f6e/activerecord/lib/active_record/connection_adapters/pool_config.rb#L83) (see also [this](https://github.com/rails/rails/blob/00cc4ff0259c0185fe08baadaa40e63ea2534f6e/activesupport/lib/active_support/fork_tracker.rb#L6-L25)) and Solid Queue is never messing with connections outside Active Record. It seems there's something else going on here, for these warnings to happen.

flavorjones:
@rosa I don't think we need to do anything here. Rails and sqlite3 do all the right things to make sure any connections carried over a `fork()` are not used. The lines you highlight make sure of that, but also [this line](https://github.com/rails/rails/blob/00cc4ff0259c0185fe08baadaa40e63ea2534f6e/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L18) is supposed to make sure the warnings aren't even emitted.

So I'm not sure how those warnings are even being emitted. @rosa have you reproduced what's described in #506?

flavorjones:
@alxckn In Rails 8, rails/rails@7c34062b contains [this line](https://github.com/rails/rails/commit/7c34062b#diff-13f18e985e903996e81f573b1fd364c19bddde279bc2797181491e6c146bcdc7R18) to suppress this warning.

What version of Rails are you seeing this with?

rosa:
> also [this line](https://github.com/rails/rails/blob/00cc4ff0259c0185fe08baadaa40e63ea2534f6e/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L18) is supposed to make sure the warnings aren't even emitted.

Ohhhh! I had forgotten about this! Thanks a lot, @flavorjones 🙏 


> So I'm not sure how those warnings are even being emitted. @rosa have you reproduced what's described in https://github.com/rails/solid_queue/issues/506?

No, I haven't! But I imagined it had to do with my super simple SQLite setup. I hadn't realised we were supressing the warnings in Rails! 😳

alxckn:
@rosa @flavorjones Thanks for looking thoroughly into this, indeed this all seems to be taken care of with active record (TIL) even though the warning only gets suppressed in rails 8.

PR #516, opened 2025-02-15
Have all processes yield self to lifecycle hooks

### LINKED ISSUE #513: Feature Request: include queues and worker index as arguments to `on_worker_start` hook
For metrics reporting this would be helpful to know, much in the same way that Puma's `on_worker_boot` hook will receive the index of the current worker.

If such a feature would be accepted I may be able to work on a PR for it this weekend.

Th3-M4jor:
Makes it so that Workers will yield self to life cycle hooks and makes some adjustments related to what's public/private in them.

I did not do this for the dispatcher or supervisor as I felt that did not make sense.

closes #513

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,6 +379,8 @@ And into two different points in the worker's, dispatcher's and scheduler's life
 - `(worker|dispatcher|scheduler)_start`: after the worker/dispatcher/scheduler has finished booting and right before it starts the polling loop or loading the recurring schedule.
 - `(worker|dispatcher|scheduler)_stop`: after receiving a signal (`TERM`, `INT` or `QUIT`) and right before starting graceful or immediate shutdown (which is just `exit!`).
 
+Each of these hooks has an instance of the supervisor/worker/dispatcher/scheduler yielded to the block so that you may read its configuration for logging or metrics reporting purposes.
+
 You can use the following methods with a block to do this:
 ```ruby
 SolidQueue.on_start
@@ -396,8 +398,20 @@ SolidQueue.on_scheduler_stop
 
 For example:
 ```ruby
-SolidQueue.on_start { start_metrics_server }
-SolidQueue.on_stop { stop_metrics_server }
+SolidQueue.on_start do |supervisor|
+  MyMetricsReporter.process_name = supervisor.name
+
+  start_metrics_server
+end
+
+SolidQueue.on_stop do |_supervisor|
+  stop_metrics_server
+end
+
+SolidQueue.on_worker_start do |worker|
+  MyMetricsReporter.process_name = worker.name
+  MyMetricsReporter.queues = worker.queues.join(',')
+end
 ```
 
 These can be called several times to add multiple hooks, but it needs to happen before Solid Queue is started. An initializer would be a good place to do this.
--- a/lib/solid_queue/scheduler.rb
+++ b/lib/solid_queue/scheduler.rb
@@ -5,7 +5,7 @@ module SolidQueue
     include Processes::Runnable
     include LifecycleHooks
 
-    attr_accessor :recurring_schedule
+    attr_reader :recurring_schedule
 
     after_boot :run_start_hooks
     after_boot :schedule_recurring_tasks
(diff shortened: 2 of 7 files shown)


lib/solid_queue/configuration.rb
```diff
@@ -33,12 +33,15 @@ def instantiate
...
 
     def configured_processes
-      if only_work? then workers
-      else
-        dispatchers + workers + schedulers
+      @configured_processes ||= begin
+        if only_work? then workers
+        else
+          dispatchers + workers + schedulers
+        end
```
Th3-M4jor, line 44: Without memoizing this, we would end up with worker ids of `3, 4, 5` instead of `1, 2, 3` in one of the modified tests. Which would be confusing.

rosa:
Hey @Th3-M4jor, thanks for this! 

However, I still don't understand the need for an incremental index for each worker 🤔  Could you elaborate more why this is necessary? These indexes would be duplicated across different supervisors (in the case you're running jobs in multiple machines). 

All processes now [have a unique generated name](https://github.com/rails/solid_queue/blob/0a6c3b69872d2ee39072b91e23ceab5327725751/lib/solid_queue/processes/base.rb#L12-L39), [which is used to locate them](https://github.com/rails/solid_queue/blob/0a6c3b69872d2ee39072b91e23ceab5327725751/lib/solid_queue/supervisor.rb#L176), and they carry other identifying information such as the hostname where they're running and their PID. Wouldn't this be more than enough to identify them?

Th3-M4jor:
For my particular case having an incremented index for each worker fits better for metrics reporting. It more closely mirrors how Puma identifies each of its forked workers, and for that I inject the k8s pod name to differentiate there.

Though I do see the point that it might not work for everyone's needs, what if I renamed `worker_id` to `worker_index` and added a `worker_name` which returned the unique generated name?

rosa:
I think you don't need a `worker_name` 🤔  You can simply access it via `#name`, no? 

About `worker_index`, if this is just for your particular case, I'd rather not add that, it should be something that's useful in general, not just for one particular way of reporting metrics. I think having the worker yielded in the lifecycle hook makes sense, but I don't see the need for having other attributes besides what is already there.

Th3-M4jor:
Fair enough, I removed all references to `worker_id` and squashed it down to a single commit. Accessing just the queues should work well enough for my case.

rosa:
Cool! I think, for simplicity, we could have all processes do this, not just workers.

Th3-M4jor:
Done, and similarly made a few additional things on each process class private if they were only accessed via tests or from within the class itself.

lib/solid_queue/scheduler.rb
```diff
@@ -24,6 +22,8 @@ def metadata
     end
 
     private
+      attr_reader :recurring_schedule
```
rosa, line 25: I'd keep this one public, ok to switch to `attr_reader`, but it doesn't need to be private.

lib/solid_queue/worker.rb
```diff
@@ -24,6 +26,8 @@ def metadata
     end
 
     private
+      attr_reader :pool
```
rosa, line 29: Same as with `recurring_schedule`, we can keep `pool` public.

README.md
```diff
@@ -396,8 +398,25 @@ SolidQueue.on_scheduler_stop
...
+  stop_metrics_server
+end
+
+SolidQueue.on_worker_start do |worker|
+  Rails.logger.info "Worker #{worker.name} started with queues: #{worker.queues.join(',')}"
+end
+
+SolidQueue.on_worker_stop do |worker|
+  Rails.logger.info "Worker #{worker.name} stopped with queues: #{worker.queues.join(',')}"
+end
```
rosa, line 419: I know this is just because these are examples, but Solid Queue's log subscriber already logs all this when processes start and stop 🤔  I'm just not sure if it's ok to add examples that would be discouraged because you'd end up with duplicated/redundant logs.

Th3-M4jor: I changed it to instead set a couple values in a somewhat contrived `MyMetricsReporter` class, which hopefully does a better job of showing what someone might want to use it for.

rosa:
Nice! Thanks a lot and sorry for the delay!

Issue #67, opened 2023-12-19
Comma separated queue names don't work


morgoth:
When specifying multiple queues as one string, ie:
```yaml
development:
  workers:
    - queues: real_time,background
```
this does not work. The workers are not taking scheduled jobs (real_time or background) at all.

It works for me, when queues are an array, ie:
```yaml
development:
  workers:
    - queues:
      - real_time
      - background
```

Should the documentation be fixed to list queues as array in YAML or maybe the logic to fix comma separated string?

rosa:
Hey @morgoth! Ahh, you're correct! We went back-and-forth on this, supporting first comma-separated values, then array, then back to comma-separate, then arrays... and I documented the wrong one in the README 😅  

I'll update the README, we went with the array in the end, so sorry about that!

PR #517, opened 2025-02-17
Fix README queues config value to array

ristovskiv:
Small PR since I have noticed [this](https://github.com/rails/solid_queue/issues/67) issue already, following by by the [fix](https://github.com/rails/solid_queue/commit/7838ad4bcb1efeac7179065d780410df31879b74), but I think this was unintentionally omitted.


rosa:
Oops, indeed! Thank you 🙏

Issue #512, opened 2025-02-07
SolidQueue crashes if database connection is lost, and takes Puma with it.


darinwilson:
We're running SolidQueue as a Puma plugin on a Rails 8 app, as our job processing load is currently quite small. 

We recently had an incident where the server running Puma temporarily lost the connection to Postgres. This caused SolidQueue to crash with this message:

```
PQconsumeInput() FATAL:  terminating connection due to administrator command (PG::
ConnectionBad)
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
```

and this in turn took down Puma:

```
Detected Solid Queue has gone away, stopping Puma...
- Gracefully stopping, waiting for requests to finish
```

I was able to reproduce this locally by shutting down Postgres after starting Rails. 

When running Rails without the SolidQueue Puma plugin, if the database goes away, Rails throws an error when it tries to do something with the database, but Puma stays up and the connections recover when the database comes back online.

If I run SolidQueue separately, via `bin/jobs`, it also crashes if the database goes away.

Obviously SolidQueue can't be expected to do much without a database, but would it be reasonable for it to behave as Rails does when the db goes offline, i.e. pause its activity and reconnect when the db is available again?

Thanks for all your work on this - SolidQueue has been a fantastic addition to Rails!

rosa:
Oh, interesting. This happens for the supervisor only, if any of the supervised processes crashes, the supervisor makes sure a new one is started 🤔 I think the supervisor would need some kind of recovery mechanism if the DB fails, but it could also crash for other reasons. I think it makes sense to do this, but I won't have time in the next couple of months at least, so if someone wants to submit a PR doing this, I'll be happy to review.

darinwilson:
Thanks for the feedback - that's good to know that it must be something at the supervisor level. 

I'll dig into the code a bit, and see if I can find a solution that might work.

asgeo1:
Had this same issue myself last week, where DB was briefly uncontactable, causing SolidQueue to shut down Puma and Rails.

What are the chances of the [PR](https://github.com/rails/solid_queue/pull/519) getting merged soon?

rosa:
Oh, I completely forgot about this one. I'll take a look at the PR.

rosa:
I'm struggling to reproduce this locally so I can test the PR and an alternative approach. In all cases both Puma and Solid Queue remain running 😕  This is quite strange. @darinwilson, you said: 

[...]

PR #519, opened 2025-02-19
Allow supervisor to recover after crash

darinwilson:
This is a first pass at a fix for #512: allowing Solid Queue to recover if the database goes offline (or if it fails for any other reason).

In the case of the database going away, there are a few possible scenarios that can cause the supervisor to fail, but the most common is:
  * worker tries and fails to poll the database
  * worker tries to deregister (via `after_shutdown` in `Registrable`) and fails
  * `Process#deregister` re-raises any exceptions that come up during deregistration, so worker crashes
  * supervisor detects the crashed process and tries to replace it, but fails
  * supervisor also inherits from `Process` so when it fails, it calls `deregister` just like the worker did
  * nothing else is around to catch the exception raised in `deregister` so Solid Queue terminates completely
  
After a restart, the maintenance tasks performed by the supervisor do a good job of cleaning up the loose ends left behind, so it seemed like the cleanest approach was just to let the supervisor crash, then spin up a new instance. This is handled by a new `Launcher` class that wraps `Supervisor#start` in a retry block with exponential backoff.

This also adds a new config parameter `max_restart_attempts` that allows the user to limit the number of restart attempts. If `nil`, it will retry forever, and if `0`, it won't try at all. (I made `0` the default since that's the current behavior.)

I tested with Postgres and MySQL, but didn't really know how to test SQLite or if it even made sense to. Again, this is just a first attempt - happy to try a different approach if this doesn't seem quite right.

Thanks!

### DIFF
--- a/lib/solid_queue/supervisor.rb
+++ b/lib/solid_queue/supervisor.rb
@@ -13,7 +13,7 @@ module SolidQueue
         configuration = Configuration.new(**options)
 
         if configuration.valid?
-          new(configuration).tap(&:start)
+          SolidQueue::Supervisor::Launcher.new(configuration).tap(&:start)
         else
           abort configuration.errors.full_messages.join("\n") + "\nExiting..."
         end
--- a/lib/solid_queue/supervisor.rb
+++ b/lib/solid_queue/supervisor.rb
@@ -13,7 +13,7 @@ module SolidQueue
         configuration = Configuration.new(**options)
 
         if configuration.valid?
-          SolidQueue::Supervisor::Launcher.new(configuration).tap(&:start)
+          Launcher.new(configuration).tap(&:start)
         else
           abort configuration.errors.full_messages.join("\n") + "\nExiting..."
         end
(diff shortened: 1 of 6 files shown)


darinwilson:
Looking into the test failures

lib/solid_queue/supervisor.rb
```diff
@@ -13,7 +13,7 @@ def start(**options)
         configuration = Configuration.new(**options)
 
         if configuration.valid?
-          new(configuration).tap(&:start)
+          SolidQueue::Supervisor::Launcher.new(configuration).tap(&:start)
```
rosa, line 16: Could this be?
```suggestion
          Launcher.new(configuration).tap(&:start)
```
Just for consistency 😊

rosa:
Hey @darinwilson, sorry for the delay and radio-silence on this one. I continued thinking about this, and I'm not convinced this is the right way to handle the problem. I think whatever process manages the supervisor should be in charge of restarting it, just in a similar way as the supervisor is in charge of restarting the workers when they crash. This means that when the Puma plugin is used, the whole thing shouldn't crash when Solid Queue stops. I'm going to close this PR, and will keep the original issue open.

darinwilson:
@rosa OK - no worries. Thanks for the feedback!

Issue #502, opened 2025-01-29
Extend documentation of concurrency controls


cedricpim:
Hello 👋 

I had a usage for concurrency controls and I was investigating how it worked and I found two pieces of information that I think could be useful in the README section because they weren't immediately clear to me.

The first piece was that the dispatcher, only releases one job with duration expired per run (so, using the default value of 5 minutes for `concurrency_maintenance_interval`, it means only one expired job is released every 5 minutes).

The second piece of information was that, on a job (with concurrency controls) completion, the `duration` of the already blocked jobs is "refreshed" (from what I was able to see, this happens because the concurrency keys to be considered for release must either have an open semaphore or no semaphore at all and, since dispatcher destroys expired semaphores, when a job finishes, the semaphore expiration time is increased).

Let me know if you want me to prepare a PR or you prefer to write this on your own words (or if I am wrong of course).

rosa:
Hey @cedricpim, sure! Feel free to open a PR with some changes to the README. 

> The first piece was that the dispatcher, only releases one job with duration expired per run

One per duration and per concurrency key, to ensure concurrency limits are still respected for the released jobs. In theory, this shouldn't need to release any jobs unless you have a job that is very long, as jobs should release the following blocked jobs when they finish, but of course, there's a race condition here: https://github.com/rails/solid_queue/issues/456. This means the dispatcher's concurrency maintenance fallback is necessary in that case until I fix that 😳 (it's a tricky one). 

> on a job (with concurrency controls) completion, the duration of the already blocked jobs is "refreshed"

Do you mean if a job finishes and un-blocks another job, then the new job renews the semaphore expiry time? Then yes, that's correct 👍

cedricpim:
@rosa sorry for the radio silence here. Just made a PR 👍

rosa:
Oops, I forgot to close this one after merging the PR.

PR #520, opened 2025-02-21
Improve README with extra info about concurrency

cedricpim:
Add two notes on README about conurrency controls and how blocked jobs are unblocked and for how long can they be blocked.

Related to what was discussed in https://github.com/rails/solid_queue/issues/502


README.md
```diff
@@ -426,7 +426,7 @@ class MyJob < ApplicationJob
...
... and we set the job as _ready_. Ready means it can be picked up by workers for execution. When the job finishes executing (be it successfully or unsuccessfully, resulting in a failed execution), we signal the semaphore and try to unblock the next job with the same key, if any. Unblocking the next job doesn't mean running that job right away, but moving it from _blocked_ to _ready_. Since something can happen that prevents the first job from releasing the semaphore and unblocking the next job (for example, someone pulling a plug in the machine where the worker is running), we have the `duration` as a failsafe. Jobs that have been blocked for more than duration are candidates to be released, but only as many of them as the concurrency rules allow, as each one would need to go through the semaphore dance check. This means that the `duration` is not really about the job that's enqueued or being run, it's about the jobs that are blocked waiting. When there are multiple jobs unblocked, it is important to note that, as a job finishes and the next jobs are unblocked, the `duration` timer for the still blocked jobs is reset (this happens indirectly via the expiration time of the semaphore).
```
rosa, line 429: I'd rephrase this, as it's not really necessary that there are multiple jobs unblocked (because concurrency limit > 1). The behaviour is always the same. 

Instead of: 
> When there are multiple jobs unblocked, it is important to note that, as a job finishes and the next jobs are unblocked, the `duration` timer for the still blocked jobs is reset (this happens indirectly via the expiration time of the semaphore).

I'd write this as:
> It's important to note that after one or more candidate jobs are unblocked (either because a job finishes or because `duration` expires and a semaphore is released), the `duration` timer for the still blocked jobs is reset. This happens indirectly via the expiration time of the semaphore, which is updated.

rosa:
Thanks a lot @cedricpim! I just left a small comment about one of the changes 🙏

cedricpim:
@rosa sorry for the delay on this. Just updated the PR.

rosa:
Nice, thanks a lot!

Issue #176, opened 2024-03-12
Discard duplicate jobs


matteeyah:
## Summary

The concurrency controls that currently exist allow blocking execution of "duplicate" jobs, or jobs with the same arguments. These jobs wait, then execute after the currently running job finishes. There's no way to discard all duplicate jobs and prevent them from running completely.

## Proposal

Add a way to discard jobs that get blocked by the concurrency controls.

rosa:
@matteeyah, I've got this on my mind as well! In the end, I'm not sure there's a version of https://github.com/basecamp/solid_queue/issues/105 that I like or that it's useful for what we need, so I thought, in the meantime, to add another parameter to `limits_concurrency` to discard instead of blocking. Something like: 
```ruby
on_conflict: :discard
```
By default, it'd be: 
```ruby
on_conflict: :block
```
which would be the current behaviour.

matteeyah:
That's exactly what I had in mind!

---

For reference - We're currently using https://github.com/veeqo/activejob-uniqueness. It serves a slightly different purpose, but it also checks for duplicate jobs.

It also has an `on_conflict` option - it supports logging a conflict, raising an error on a conflict and passing a proc to handle the conflict manually.

nhorton:
Another vote for following the activejob-uniqueness conventions.

I would note that of the strategies offered by the gem, `until_and_while_executing` is the one we have found most-useful.

nhorton:
Is there any more word on this? Or any openness to others trying to contribute?

rosa:
Hey @nhorton, sorry for the delay! I'm back at this now and would like to get this feature ready for version 1.0. 

> of the strategies offered by the gem, `until_and_while_executing` is the one we have found most-useful.

I think that would be the only strategy supported because of the way concurrency controls work. Right now jobs don't unblock other jobs or release the semaphore until they complete, and for simplicity, it'd be much easier to keep it that way. 

> Or any openness to others trying to contribute?

Definitely! 🙌

bilby91:
@rosa Is there an ETA for 1.0 ?

rosa:
@bilby91, my hope is by the end of August but you know, there are always unexpected things 😅 😅

bilby91:
@rosa Awesome! I look forward to testing this feature.

rosa:
In the end I won't be able to get this one in for v1.0, but I hope to get it there shortly after. Sorry!

wheee:
[...]

PR #523, opened 2025-02-25
Add capability to discard duplicate jobs with concurrency configuration

joelzwarrington:
Closes https://github.com/rails/solid_queue/issues/176

Add ability to discard duplicate jobs instead of becoming blocked. When configuring `concurrency_limits` with `on_duplicate: discard` jobs scheduled above the concurrency limit are discarded and not executed.

## Implementation Details

* adds class_attribute `concurrency_on_duplicate` and `on_duplicate` option to `concurrency_limits`
* adds methods to Semaphore to detect when a job would be above the concurrency limit
* adds logic to discard a job on dispatch when above concurrency limit

### DIFF
--- a/app/models/solid_queue/job/executable.rb
+++ b/app/models/solid_queue/job/executable.rb
@@ -65,7 +65,8 @@ module SolidQueue
       end
 
       def dispatch
-        if acquire_concurrency_lock then ready
+        if discard_on_duplicate? then discard
+        elsif acquire_concurrency_lock then ready
         else
           block
         end
--- a/app/models/solid_queue/semaphore.rb
+++ b/app/models/solid_queue/semaphore.rb
@@ -10,6 +10,10 @@ module SolidQueue
         Proxy.new(job).wait
       end
 
+      def at_limit?(job)
+        Proxy.new(job).at_limit?
+      end
+
       def signal(job)
         Proxy.new(job).signal
       end
@@ -37,6 +41,14 @@ module SolidQueue
 
       def initialize(job)
         @job = job
+      end
+
+      def at_limit?
+        if semaphore = Semaphore.find_by(key: key)
+          semaphore.value.zero?
+        else
+          false
+        end
       end
 
       def wait
--- a/lib/active_job/concurrency_controls.rb
+++ b/lib/active_job/concurrency_controls.rb
@@ -11,15 +11,17 @@ module ActiveJob
       class_attribute :concurrency_group, default: DEFAULT_CONCURRENCY_GROUP, instance_accessor: false
 
       class_attribute :concurrency_limit
+      class_attribute :concurrency_on_duplicate
       class_attribute :concurrency_duration, default: SolidQueue.default_concurrency_control_period
     end
 
     class_methods do
-      def limits_concurrency(key:, to: 1, group: DEFAULT_CONCURRENCY_GROUP, duration: SolidQueue.default_concurrency_control_period)
+      def limits_concurrency(key:, to: 1, group: DEFAULT_CONCURRENCY_GROUP, duration: SolidQueue.default_concurrency_control_period, on_duplicate: :block)
         self.concurrency_key = key
         self.concurrency_limit = to
         self.concurrency_group = group
         self.concurrency_duration = duration
+        self.concurrency_on_duplicate = on_duplicate
       end
     end
 
--- a/app/models/solid_queue/job/concurrency_controls.rb
+++ b/app/models/solid_queue/job/concurrency_controls.rb
@@ -8,7 +8,7 @@ module SolidQueue
       included do
         has_one :blocked_execution
 
-        delegate :concurrency_limit, :concurrency_duration, to: :job_class
+        delegate :concurrency_limit, :concurrency_on_duplicate, :concurrency_duration, to: :job_class
 
         before_destroy :unblock_next_blocked_job, if: -> { concurrency_limited? && ready? }
       end
@@ -34,6 +34,14 @@ module SolidQueue
       end
 
       private
+        def duplicate?
+          Semaphore.at_limit?(self)
+        end
+
+        def discard_on_duplicate?
+          concurrency_on_duplicate == :discard && duplicate?
+        end
+
         def acquire_concurrency_lock
           return true unless concurrency_limited?
 
--- a/test/models/solid_queue/job_test.rb
+++ b/test/models/solid_queue/job_test.rb
@@ -10,6 +10,14 @@ class SolidQueue::JobTest < ActiveSupport::TestCase
     end
   end
 
+  class DiscardedNonOverlappingJob < NonOverlappingJob
+    limits_concurrency key: ->(job_result, **) { job_result }, on_duplicate: :discard
+  end
+
+  class DiscardedOverlappingJob < NonOverlappingJob
+    limits_concurrency to: 2, key: ->(job_result, **) { job_result }, on_duplicate: :discard
+  end
+
   class NonOverlappingGroupedJob1 < NonOverlappingJob
     limits_concurrency key: ->(job_result, **) { job_result }, group: "MyGroup"
   end
@@ -96,6 +104,40 @@ class SolidQueue::JobTest < ActiveSupport::TestCase
     job = SolidQueue::Job.last
     assert_equal active_job.concurrency_limit, job.concurrency_limit
     assert_equal active_job.concurrency_key, job.concurrency_key
+  end
+
+  test "enqueue jobs with discarding concurrency controls" do
+    assert_ready do
+      active_job = DiscardedNonOverlappingJob.perform_later(@result, name: "A")
+      assert_equal 1, active_job.concurrency_limit
+      assert_equal "SolidQueue::JobTest::DiscardedNonOverlappingJob/JobResult/#{@result.id}", active_job.concurrency_key
+    end
+
+    assert_discarded do
+      active_job = DiscardedNonOverlappingJob.perform_later(@result, name: "A")
+      assert_equal 1, active_job.concurrency_limit
+      assert_equal "SolidQueue::JobTest::DiscardedNonOverlappingJob/JobResult/#{@result.id}", active_job.concurrency_key
+    end
+  end
+
+  test "enqueue jobs with discarding concurrency controls when below limit" do
+    assert_ready do
+      active_job = DiscardedOverlappingJob.perform_later(@result, name: "A")
+      assert_equal 2, active_job.concurrency_limit
+      assert_equal "SolidQueue::JobTest::DiscardedOverlappingJob/JobResult/#{@result.id}", active_job.concurrency_key
+    end
+
+    assert_ready do
+      active_job = DiscardedOverlappingJob.perform_later(@result, name: "A")
+      assert_equal 2, active_job.concurrency_limit
+      assert_equal "SolidQueue::JobTest::DiscardedOverlappingJob/JobResult/#{@result.id}", active_job.concurrency_key
+    end
+
+    assert_discarded do
+      active_job = DiscardedOverlappingJob.perform_later(@result, name: "A")
+      assert_equal 2, active_job.concurrency_limit
+      assert_equal "SolidQueue::JobTest::DiscardedOverlappingJob/JobResult/#{@result.id}", active_job.concurrency_key
+    end
   end
 
   test "enqueue jobs with concurrency controls in the same concurrency group" do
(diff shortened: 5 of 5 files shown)


app/models/solid_queue/job/executable.rb
```diff
@@ -66,6 +66,8 @@ def prepare_for_execution
 
       def dispatch
         if acquire_concurrency_lock then ready
+        elsif discard_concurrent?
+          discard
```
rosa, line 70: I don't think this is the way 🤔  We're in the middle of a transaction here, and the job hasn't even been committed to the DB. It makes no sense to delete a record in the same transaction you're creating it. It'd make sense to roll that transaction back instead.

app/models/solid_queue/semaphore.rb
```diff
@@ -39,6 +43,14 @@ def initialize(job)
         @job = job
       end
 
+      def at_limit?
+        if semaphore = Semaphore.find_by(key: key)
+          semaphore.value.zero?
```
rosa, line 48: This is vulnerable to race conditions and the reason we don't check it in this way when blocking jobs. If two concurrent jobs are claiming the semaphore, both of them will see it open, and none will be discarded. Then, both will run together because we won't block them either.

joelzwarrington: You raise an excellent point. Do you have a suggestion to how we might be able to avoid this race condition?

Are you thinking [Pessimistic Locking](https://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html) might help?

joelzwarrington:
Thanks for the review @rosa ❤️, I'll review your comments more thoroughly and make some changes. 😊

app/models/solid_queue/job/executable.rb
joelzwarrington: Perhaps I don't fully understand how we're getting into this code path, but from my investigation, it looks as though there isn't an open transaction at this time.

I tried the following and running the test suite and didn't hit any open transactions.

```
raise "open transactions" if ApplicationRecord.connection.open_transactions.positive?
```

I still believe we'd want to discard here. Let me know your thoughts.

app/models/solid_queue/job.rb
```diff
@@ -10,19 +10,22 @@ class EnqueueError < StandardError; end
...
-            enqueued_jobs.each do |enqueued_job|
-              active_jobs_by_job_id[enqueued_job.active_job_id].provider_job_id = enqueued_job.id
-              active_jobs_by_job_id[enqueued_job.active_job_id].successfully_enqueued = true
-            end
+          enqueued_jobs_by_active_job_id = prepare_all_for_execution(jobs).index_by(&:active_job_id)
+
+          active_jobs.each do |active_job|
+            job = enqueued_jobs_by_active_job_id[active_job.job_id]
+            active_job.provider_job_id = job&.id
+            active_job.successfully_enqueued = job.present?
```
joelzwarrington, line 22: I've reorganized this method so that we can efficiently set `successfully_enqueued` for both enqueued/blocked and discarded jobs.

app/models/solid_queue/job.rb
```diff
@@ -49,7 +52,7 @@ def create_from_active_job(active_job)
         def create_all_from_active_jobs(active_jobs)
           job_rows = active_jobs.map { |job| attributes_from_active_job(job) }
           insert_all(job_rows)
-          where(active_job_id: active_jobs.map(&:job_id))
+          where(active_job_id: active_jobs.map(&:job_id)).order(id: :asc)
```
joelzwarrington, line 55: I think this may have been an unintentional bug, or would be something that's good to do moving forward with more concurrency configuration.

Previous to adding this order, the jobs could be dispatched out of order they were provided in. This isn't really a problem when there isn't concurrency controls on the job, but when using concurrency controls, it might have unintended consequences.

So if you were to do:

```ruby
jobs = [
  job_1 = DiscardedNonOverlappingJob.new(@result, name: "A"),
  job_2 = DiscardedNonOverlappingJob.new(@result, name: "B")
]

enqueued_jobs_count = SolidQueue::Job.enqueue_all(jobs)
```

There wasn't any guarantee that job_1 would be dispatched before job_2. Now, it should be as these jobs are bulk inserted with auto-incrementing IDs, job_1 will have the lower id.

app/models/solid_queue/job/executable.rb
```diff
@@ -66,6 +66,7 @@ def prepare_for_execution
 
       def dispatch
         if acquire_concurrency_lock then ready
+        elsif discard_concurrent? then discard
```
joelzwarrington, line 69: I don't believe discard does anything here, since it just deletes the execution. But I believe in this point we won't have an execution?

This isn't deleting the job as I once thought it was. This is effectively doing 'nothing'.

joelzwarrington (commented):
Hi @rosa, I've made a few changes and believe to have encountered weird interactions with bulk enqueuing which might be a bug. Let me know if you'd like those changes split out into another PR.

rajeevriitm:
I am really in need of this feature. As its not yet available , could you help me in implementing this ? I have a `ContinuousSearchJob` thats run with `recurring.yml` . Sometimes the job takes longer and I dont want another job to be added to the queue unless the existing one is finished. Could you help me with the best way I can implement this currently?

app/models/solid_queue/job/executable.rb
rosa: > I don't believe discard does anything here, since it just deletes the execution. But I believe in this point we won't have an execution?

That's right, we won't have an execution yet. However, we need to ensure we don't create the job. I think this will result in the job record being created. That's why I think we need to rollback the current transaction. There should be a transaction open since this is called as an `after_create` callback on the job.

app/models/solid_queue/job.rb
rosa: 🤔  I think the previous implementation did that as well (only setting `successfully_enqueued = true` for jobs that had been actually enqueued), it should work without doing anything for discarded jobs, as long as discarded jobs don't create a job record. I must be missing something here, why do you think the previous implementation didn't set that correctly?

joelzwarrington: @rosa previous implementation would only set successfully_enqueued = true, but wouldn't set successfully_enqueued = false if it wasn't enqueued.

If you're okay with that quirk, then I can revert this change.

app/models/solid_queue/job.rb
joelzwarrington: Intuitively, if I were using bulk enqueue and concurrency controls, I would want it to enqueue the jobs in the order they were created, no?

If you don't want to make that guarantee, I can remove these changes and instead add a more prominent call-out in the bulk enqueue documentation that ordering is not guaranteed.

app/models/solid_queue/job/executable.rb
joelzwarrington: @rosa ah - thanks for pointing out the callback. Because `dispatch` is also used in bulk enqueueing, should we rollback all jobs in the bulk enqueue if one failed to be enqueued due to concurrency controls?

joelzwarrington:
@rajeevriitm if you have a look at the read me you should find concurrency controls are available for you to use.

The only difference being that jobs will always be blocked, so you may get several jobs that are awaiting execution.

joelzwarrington:
@rosa any updates on your review here?

rynonl:
@rosa Just pinging this. Would love to see this functionality get merged!

danlo:
@rosa +1

rosa:
Hey, sorry everyone! I've been crazy busy at work in a bunch of unrelated things, but I'm planning for at least a day of Solid Queue work to catch up with all issues and PRs very soon. Thanks y'all for your patience and sorry for the delay!

app/models/solid_queue/job/executable.rb
rosa: > Because dispatch is also used in bulk enqueueing, should we rollback all jobs in the bulk enqueue if one failed to be enqueued due to concurrency controls?

Hmm... tricky one to support... No, I think we shouldn't do that 🤔  I think we'd have to split the bulk enqueuing into two: the jobs that are concurrency controlled on their own, and then the jobs that are not. Then the ones that are concurrency controlled would have to be enqueued one by one. This should be documented in the README, warning that using `peform_all_later` with concurrency-limited jobs is not going to get any performance advantages at all.

joelzwarrington:
Thanks for the review. Gonna have a look and implement any feedback this week

lib/active_job/concurrency_controls.rb
```diff
@@ -11,15 +11,17 @@ module ConcurrencyControls
       class_attribute :concurrency_group, default: DEFAULT_CONCURRENCY_GROUP, instance_accessor: false
 
       class_attribute :concurrency_limit
+      class_attribute :concurrency_at_limit
```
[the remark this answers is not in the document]
joelzwarrington: Resolved via f7a445069cbdd78842ca519e5507decb0d957fa8

joelzwarrington:
It'd be a good idea to update https://github.com/rails/rails/blob/c049370882474acffbc8fd5be458e6ecd1d7d1cb/guides/source/active_job_basics.md?plain=1#L346 when merging this as I've changed `SolidQueue::Job::EnqueueError` to inherit from `ActiveJob::EnqueueError`.

app/models/solid_queue/job/executable.rb
joelzwarrington: Fixed in 88e9ef6.

I've updated discarding logic to work in both single and bulk enqueuing use cases.

* When enqueuing a single job which conflicts with concurrency configuration, a `SolidQueue::Job::EnqueueError` is raised and the `SolidQueue::Job` creation is rolledback automatically and no `SolidQueue::Job` is created
  * `SolidQueue::Job::EnqueueError` was changed to inherit from `ActiveJob::EnqueueError` which better integrates with ActiveJob framework. ActiveJob is [configured to rescue](https://github.com/rails/rails/blob/c049370882474acffbc8fd5be458e6ecd1d7d1cb/activejob/lib/active_job/enqueuing.rb#L140) from this error and keeps `successfully_enqueued?` as false and sets the `enqueue_error`
* When bulk enqueuing jobs, jobs are inserted all at once, but then dispatched one by one. When a job is dispatched and conflicts with concurrent jobs, it's discarded (and destroyed)
* When a job is scheduled to enqueue in the future, it enqueues regularly and successfully, but when it's dispatched and would otherwise conflict with concurrent jobs it's discarded (and destroyed)

joelzwarrington:
@rosa, pull request is ready for you review again! Let me know what you think.

If you're on-board with these changes, I'll also add a PR to update the RoR edge guide.

README.md
```diff
@@ -426,18 +426,21 @@ In the case of recurring tasks, if such error is raised when enqueuing the job c
 
 ## Concurrency controls
 
-Solid Queue extends Active Job with concurrency controls, that allows you to limit how many jobs of a certain type or with certain arguments can run at the same time. When limited in this way, jobs will be blocked from running, and they'll stay blocked until another job finishes and unblocks them, or after the set expiry time (concurrency limit's _duration_) elapses. Jobs are never discarded or lost, only blocked.
+Solid Queue extends Active Job with concurrency controls, that allows you to limit how many jobs of a certain type or with certain arguments can run at the same time. When limited in this way, jobs will be blocked from running, and they'll stay blocked until another job finishes and unblocks them, or after the set expiry time (concurrency limit's _duration_) elapses. Jobs can can be configured to either be discarded or blocked.
```
Gnirt, line 429:
```suggestion
Solid Queue extends Active Job with concurrency controls, that allows you to limit how many jobs of a certain type or with certain arguments can run at the same time. When limited in this way, jobs will be blocked from running, and they'll stay blocked until another job finishes and unblocks them, or after the set expiry time (concurrency limit's _duration_) elapses. Jobs can be configured to either be discarded or blocked.
```

marelons1337:
Would love to see that implemented, defo +1

joelzwarrington:
👋 @rosa any idea when I should expect a re-review? I don't mind if you go a different direction but would love to get this type of functionality implemented

rosa:
@joelzwarrington sorry for the delay 😓 I've been doing this in my free time and didn't have time over the weekend. I hope to get to it sometime this week or next week, when I'll have a few hours dedicated to Solid Queue 🤞  It's top of my list for SQ work. I'd like to incorporate the idea from #586 because I think that simplifies a lot, but I need to do that carefully to ensure bulk enqueuing works as it should.

rosa:
Cherry-picked a bunch of commits from here and created https://github.com/rails/solid_queue/pull/594, so I'm closing this one. Thanks a lot! 🙏

PR #529, opened 2025-03-07
Pass job.id to the executing job as provider_job_id for access within perform()

### LINKED ISSUE #514: Accessing the job from within the perform method
I'm currently using `DelayedJob` with this custom plugin that allows me to access the `provider_job_id` from within the `perform` method of a job:

```
class Delayed::ActiveJobPlugin < Delayed::Plugin
  callbacks do |lifecycle|
    lifecycle.before(:invoke_job) do |job|
      job.payload_object.job_data['provider_job_id'] = job.id if job.payload_object.respond_to?(:job_data)
    end
  end
end

Delayed::Worker.plugins << Delayed::ActiveJobPlugin
```

The Sidekiq adapter in Rails also does something similar:

https://github.com/rails/rails/pull/25961/files#diff-e1ee600ca5fd100da20810ef5acbab89546064fc323c7b1e6bdb6ed5e681a9d3L40

I'm intending to migrate to Solid Queue but just wondered if this is an option currently? From looking at the source I don't see the job id being passed in. I have two use cases:

- Keeping track of a job that generates another job at the end of its run (so a kind of recurring job configurable by the user). I store the `job_id` in the database so that I can destroy that job and schedule a new job if the user changes the time it should run.
- Skipping the execution of a job if the object passed into the job has an `updated_at` that is greater than the job's `created_at`. This signals that there's now a newer job scheduled that should take precedence. I use this for WebPush notifications.

brendon:
Sometimes it's handy to have `provider_job_id` available within `perform`. The Sidekiq adapter does it and Rails allows for it:

https://github.com/rails/rails/blob/aa2bfad1a137c9add1b35de118b391494e13ca06/activejob/lib/active_job/core.rb#L35

So far I'm not sure how to test this and the suite doesn't appear to go so far as testing the execution of jobs or at least what is available within `perform()`. Some guidance would be much appreciated :)

Closes #514

### DIFF
--- a/app/models/solid_queue/claimed_execution.rb
+++ b/app/models/solid_queue/claimed_execution.rb
@@ -92,7 +92,7 @@ class SolidQueue::ClaimedExecution < SolidQueue::Execution
 
   private
     def execute
-      ActiveJob::Base.execute(job.arguments)
+      ActiveJob::Base.execute(job.arguments.merge(provider_job_id: job.id))
       Result.new(true, nil)
     rescue Exception => e
       Result.new(false, e)
--- a/app/models/solid_queue/claimed_execution.rb
+++ b/app/models/solid_queue/claimed_execution.rb
@@ -92,7 +92,7 @@ class SolidQueue::ClaimedExecution < SolidQueue::Execution
 
   private
     def execute
-      ActiveJob::Base.execute(job.arguments.merge(provider_job_id: job.id))
+      ActiveJob::Base.execute(job.arguments.merge("provider_job_id" => job.id))
       Result.new(true, nil)
     rescue Exception => e
       Result.new(false, e)
--- a/test/dummy/app/jobs/provider_job_id_job.rb
+++ b/test/dummy/app/jobs/provider_job_id_job.rb
@@ -1,0 +1,5 @@
+class ProviderJobIdJob < ApplicationJob
+  def perform
+    JobBuffer.add "provider_job_id: #{provider_job_id}"
+  end
+end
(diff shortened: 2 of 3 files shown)


brendon:
In my quick testing on my local app with this branch it still doesn't seem to work, so please hold off merging :)

brendon:
Ok, now it's working. My mistake was passing a symbol instead of a string. A test would still be good though.

rosa:
Thank you @brendon! 

> So far I'm not sure how to test this and the suite doesn't appear to go so far as testing the execution of jobs or at least what is available within perform(). Some guidance would be much appreciated :)

Good question! I think you could try adding a new test job (to `test/dummy/app/jobs`) that does something with the `provider_job_id`, in a way you can check it later (perhaps in a similar way to `AddToBufferJob`. I'd probably include this test in `test/models/solid_queue/claimed_execution_test.rb`.

brendon:
Thanks for the tip. I've added a passing test. I think this one is ready to go.

I'm not 100% sure that's the best point to inject into the arguments but I couldn't find a better spot.

brendon:
Hi @rosa, did you need anything more from me to merge this PR? :)

rosa:
Hey @brendon, sorry for the delay! I think this is ready, thank you so much! 🙏

brendon:
No worries :) Thank you too.

If it’s not too much trouble would you be able to release a new gem version in the next little while? :)

rosa:
@brendon yes! Going to release a new version today.

brendon:
Nice! :D Thank you :)

PR #536, opened 2025-03-19
Add solid_queue_monitor as a Recommended Monitoring Option in Readme file

vishaltps:
## Overview
This PR adds [solid_queue_monitor](https://github.com/vishaltps/solid_queue_monitor) as a recommended monitoring solution for Solid Queue, with particular emphasis on its benefits for Rails API-only applications and projects requiring minimal dependencies.

## Changes
- Added a dedicated "Monitoring" section to the table of contents
- Created a comprehensive Monitoring section that showcases both monitoring options:
  - **solid_queue_monitor**: Positioned as the first option, highlighting its zero-dependency nature and suitability for API-only applications
  - **mission_control-jobs**: Maintained as the officially supported dashboard with its comprehensive feature set
- Added a reference to solid_queue_monitor in the "Failed jobs and retries" section as an alternative to mission_control-jobs

## Motivation
While mission_control-jobs is the officially supported monitoring solution, it may not be the best fit for all applications, particularly:
- Rails API-only applications where adding Rails views/assets dependencies is unnecessary
- Projects where a lightweight, zero-dependency solution is preferred
- Applications where a simpler UI with core monitoring features is sufficient

The solid_queue_monitor gem fulfills these needs while offering all essential monitoring features, making it an excellent complement to the existing mission_control-jobs option.

## Benefits
This change improves the documentation by:
- Providing options that cater to different application architectures and requirements
- Helping developers make informed choices about monitoring tools based on their specific needs
- Highlighting community contributions that enhance the Solid Queue ecosystem

By offering multiple monitoring options, we accommodate different user preferences and application constraints, making Solid Queue more versatile and accessible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,12 +25,12 @@ Solid Queue can be used with SQL databases such as MySQL, PostgreSQL or SQLite, 
 - [Concurrency controls](#concurrency-controls)
 - [Failed jobs and retries](#failed-jobs-and-retries)
   - [Error reporting on jobs](#error-reporting-on-jobs)
+- [Monitoring](#monitoring)
 - [Puma plugin](#puma-plugin)
 - [Jobs and transactional integrity](#jobs-and-transactional-integrity)
 - [Recurring tasks](#recurring-tasks)
 - [Inspiration](#inspiration)
 - [License](#license)
-
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


rosa:
Hey @vishaltps, thanks for this, and thanks for your work on `solid_queue_monitor`! 
Currently, I'm not comfortable recommending an option to manage jobs (as the first option, also) over which I have no control and that uses private APIs (basically going to low-level SQ data structures) that can change at any moment, as it could cause problems I wouldn't be best equipped to help with. 

Just a note: [Mission Control can be used by API-only apps](https://github.com/rails/mission_control-jobs?tab=readme-ov-file#api-only-apps-or-apps-using-vite_rails-and-other-asset-pipelines-outside-rails) so that shouldn't be a blocker for people running one of such apps.

Issue #546, opened 2025-04-04
When the worker dies, the concurrency limit is not reset


OlegChuev:
I have a job that can run for an extended period — let’s call it `InfiniteSleepJob`:

```rb
class InfiniteSleepJob < ApplicationJob
  # limit only one simultaneous execution of operation per key
  limits_concurrency to: 1, key: ->(key, *_args) { key },  group: "my_group", duration: 1.day

  def perform(*args)
    # do smt
  end
end
```

When I add a few jobs to the queue and the worker unexpectedly dies during execution (e.g., due to an HA event or if I forcibly kill the Docker container with docker kill container), the concurrency limit gets stuck, even after the job is pruned.

Here’s an example of a job that was pruned, but the concurrency limit wasn’t removed:
![Image](https://github.com/user-attachments/assets/3b0f0bc8-1200-4a5c-86ee-ca869ccb170e)

![Image](https://github.com/user-attachments/assets/732eccd8-0200-41fc-add7-9c0da937ccdb)

Here’s my `queue.yml`:

```yml
default: &default
  dispatchers:
    - polling_interval: 1
      batch_size: 500
      concurrency_maintenance_interval:  120
  workers:
    - queues: "default"
      threads: 3
      processes: 1
      polling_interval: 1
```

I’d expect the queue to unlock as soon as the job fails, allowing the blocked jobs to proceed as expected.

Could you kindly clarify whether this is the expected behavior? Am I missing something?

rosa:
Hey @OlegChuev, yeah, it's expected behaviour if the worker dies unceremoniously (eg. a `kill -9` signal). It wouldn't have a chance to clean up after itself, which includes the concurrency limit. The duration parameter determines how long the system will wait to unblock jobs that didn't get successfully unblocked for any reason, so if your long running job starts at, say 17:00, another one gets enqueued and blocked at 18:00, and the first one dies at 20:00, it won't be until the following day at 17:00 (`1.day` duration) that the blocked job will be considered to be unlocked.

rosa:
Although I think I could improve this case, when releasing claimed jobs whose worker got killed... 🤔 I'll work on this next week.

PR #547, opened 2025-04-08
Release concurrency locks when supervisor fails the jobs handled by workers

nerisa:
When workers 'die' because of various reasons, the supervisor unlocks the jobs that were being handled by those workers by deleting the `claimed_executions` for the job and creating `failed_executions`. However, it does not release any concurrency locks taken by the jobs. This can be an issue if the concurrency limit was reached for the job because no further jobs can run until the concurrency duration is reached.

In this PR, I call `job.unblock_next_blocked_job` when the supervisor releases all claimed executions of the worker.

Fixes https://github.com/rails/solid_queue/issues/546

### DIFF
--- a/app/models/solid_queue/claimed_execution.rb
+++ b/app/models/solid_queue/claimed_execution.rb
@@ -39,7 +39,10 @@ class SolidQueue::ClaimedExecution < SolidQueue::Execution
     def fail_all_with(error)
       SolidQueue.instrument(:fail_many_claimed) do |payload|
         includes(:job).tap do |executions|
-          executions.each { |execution| execution.failed_with(error) }
+          executions.each do |execution|
+            execution.failed_with(error)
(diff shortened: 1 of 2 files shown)


nerisa:
I couldnt figure out why `test/integration/processes_lifecycle_test.rb:171 ` fails. If I run the test file independently, it passes but when I run the whole test suite, I get the same error as in the CI.

rosa (approved):
Hey @nerisa, thanks a lot for submitting this PR and sorry for the delay! This looks great. I think the failure is due to that test being flaky in some circumstances. I've got it pending to make them completely stable 😳

p-schlickmann:
Hi @rosa 👋

I’d love to help stabilize the tests.  I understand they are tricky due to their concurrent nature

Do you have any previous PRs or attemps that didn't work out, so I can use them as reference?

Issue #186, opened 2024-03-23
Dynamic scheduled tasks


wollistik:
Hi @rosa,
I really appreciate the work you have done here and I am eager to switch over to solid_queue. There is only one thing left, which is holding me back.

Other gems like resque-scheduler or sidekiq-scheduler offer the ability to dynamically add or remove tasks to the schedule (see https://github.com/resque/resque-scheduler?tab=readme-ov-file#dynamic-schedules). Since everything seems to be stored in the database for solid_queue, this should be quite easy to achieve. 

Maybe it is only a documentation issue and it is already possible (this would be awesome 🤞 ).
Happy to hear from you!

rosa:
@wollistik, no, this is not supported yet, you didn't miss it 😅 

This is something we can consider adding for sure. May I ask what's your use case for it?

wollistik:
Hi @rosa,
we have some cron like jobs, where the user is able to

* Disable/Enable running the job
* Change the schedule on when and how often the job is performed

Therefore the dynamic scheduling feature was really nice to implement these requirements.

But after I thought about this the last two days, I might be completely redesign this feature and come up with a different solution, because it was rarely used and still required some quirks to get it working.

rosa:
Ahh got it! That makes sense 👍  Yes, I thought a bit more about it yesterday and this is something I want to implement eventually 😊

abrunner94:
Maybe this helps, but the way I implemented dynamic cron schedules is by having one job run every minute and check if the cron schedule matches the one stored in my database table, with my table containing cron expressions. I use Fugit to check if the cron matches the current time.

AquisTech:
@abrunner94 Even I have done a similar thing. Mine is not exactly the cron like scheduling but I am having different schedules.
First task is scheduled to be executed after 15 days and then onwards every alternate days.
So in the execution of first job I enqueue next job and in that job I keep enqueueing next jobs every 2 days as per condition.

KellyBennett:
I will just add my use case as well in case it helps motivate the feature!

TL;DR: I empowered our product managers to be able to control schedules without needing to recruit an engineer and get a new version of the app deployed.


More details;

[...]

PR #553, opened 2025-04-23
Dynamic scheduled tasks

cupatea:
This pull request closes [rails/solid_queue#186] and brings in a resque-scheduler style dynamic schedules feature, allowing you to add or remove recurring tasks at runtime without touching your static config file.

**What’s new:**

- Dynamic scope (static: false) on `SolidQueue::RecurringTask` model.
- Renamed methods/vars in `SolidQueue::Scheduler::RecurringSchedule` to distinguish static vs. dynamic schedules.
- On boot, `@configured_tasks` now includes both static and dynamic tasks.
- Added `SolidQueue::Scheduler::RecurringSchedule.update_scheduled_tasks`:
  - Schedules any new dynamic tasks added since the last run.
  - Unschedules any dynamic tasks removed from the database.
  - Refreshes @configured_tasks to match the current DB state.
- `SolidQueue::Configuration` no longer requires a non-blank static config file - pure dynamic scheduling is now supported.
- `SolidQueue::Scheduler` watches for changes after launch and updates its metadata so the running process always reflects the true set of recurring tasks.

Tests verify that adding or dropping dynamic tasks at runtime correctly updates what’s scheduled.


cupatea:
Hi @rosa 👋, could you please take a look at this PR when you have a moment? Thanks so much!

rosa:
Hey @cupatea, thanks for this! It's a good start, but it needs a few changes. The main ones are: 
- We need to make the sleep interval for the scheduler configurable, as a sort of polling interval, to reload dynamic tasks. 
- The way to use this needs to be a bit friendlier, without users having to go and figure out the details to create a `SolidQueue::RecurringTask` record. Something like `SolidQueue.schedule_recurring_task` could work, and it needs to be documented. 
- We'd need the ability to delete dynamically scheduled tasks as well, by task ID. If the task is static, this could simply raise an error. 

And then some other more specific changes that I'll note in the code.

cupatea:
Hi @rosa, thank you for the feedback! I think it's ready for the second round of review

bruno-berchielli:
@cupatea , in the docs files you mention `SolidQueue.delete_recurring_task` but I think it should be `SolidQueue.destroy_recurring_task`, also, you might consider using the `key` instead of `id`, because that's is what probably gonna be easily available on the app level, no?
`SolidQueue::RecurringTask.find_by(key: solid_queue_key)&.destroy!`

cupatea:
> @cupatea , in the docs files you mention `SolidQueue.delete_recurring_task` but I think it should be `SolidQueue.destroy_recurring_task`, also, you might consider using the `key` instead of `id`, because that's is what probably gonna be easily available on the app level, no? `SolidQueue::RecurringTask.find_by(key: solid_queue_key)&.destroy!`

Thanks for noticing, on a way to fix that! 
And yes, using a key instead of ID makes more sense to me as well. To prevent deleting a static task, this method will raise a record not found error in case of specifying a static task key (or if you try deleting a task that does not exist)

rosa:
Hey @cupatea, sorry for the delay here. I'm finally working towards Solid Queue 2.0 with the plan of including this. I've pushed a bunch of changes. Some are stylistic and simplifications, but I've also changed the default behaviour. I think this feature should be opt-in, disabled by default, as people might not need dynamic tasks at all and could save the extra queries to the DB.

cupatea:
> Hey @cupatea, sorry for the delay here. I'm finally working towards Solid Queue 2.0 with the plan of including this. I've pushed a bunch of changes. Some are stylistic and simplifications, but I've also changed the default behaviour. I think this feature should be opt-in, disabled by default, as people might not need dynamic tasks at all and could save the extra queries to the DB.

Thank you!

I noticed something in `app/models/solid_queue/recurring_task.rb:40`.
It looks like this should use `**options.merge(static: false)` instead of `reverse_merge`, since we want to ensure the static option is always overridden

rosa:
> I noticed something in app/models/solid_queue/recurring_task.rb:40.
It looks like this should use `**options.merge(static: false)` instead of `reverse_merge`, since we want to ensure the > static option is always overridden

Ohh, totally! My bad. I had it backwards before, switched it around, but forgot to update the method. Thanks!

rosa:
This should be ready to go but I'd like to test it a bit better before merging. Have you been using this already in production by any chance, @cupatea @lrjbrual?

cupatea:
> This should be ready to go but I'd like to test it a bit better before merging. Have you been using this already in production by any chance, @cupatea @lrjbrual?

Not really - I migrated from Resque to this and deployed to the test server, but I haven’t deployed it to production yet.
Our production is pretty light in terms of jobs number, so it's more for getting rid of Redis dependency

rosa:
No problem, let me test this a bit more in production in case I find something, but I should merge this next week.

rosa:
Sorry for the delay! I was derailed with other priorities at work and couldn't ship this. I've done that now so I'll merge this at the end of today and will release a new version.

PR #561, opened 2025-05-10
Document adapter behavior with recurring tasks

julik:
At the moment it is not possible to use solid_queue for scheduling while enqueueing the scheduled job into a different adapter. This should, at least, be documented.


rosa:
Hey @julik, thank you! But [this is already documented](https://github.com/rails/solid_queue?tab=readme-ov-file#recurring-tasks): 
 
Screenshot 2025-05-10 at 19 20 33

julik:
Oh! missed it completely! thank you.On 10 May 2025, at 19:21, Rosa Gutierrez ***@***.***> wrote:Hey @julik, thank you! But this is already documented:

Issue #560, opened 2025-05-07
Suggestion: Should preserve_finished_jobs really default to true?


stefanvermaas:
We've just hit a tricky issue that I think many teams might run into, especially when adopting solid_queue in smaller projects without deep background job needs.

By default, SolidQueue preserves all finished jobs ([source](https://github.com/rails/solid_queue/blob/0ee21e49de3bbadf1c72294cdbadf12978486d6c/lib/solid_queue.rb#L40)). That feels like a safe option at first glance, but if you're not aware of this setting, you might only discover it once it's too late.

In our case, we had a small Rails app using `SolidQueue` out of the box. It was deployed with Kamal, everything worked perfectly for months. Until one day, the app became unreachable. No traffic, no response, and only a single error message from hours earlier: "could not connect to database."

The database server was up. The app server was up. But after digging in, we discovered that the entire server had run out of disk space.

The root cause? All those finished background jobs were quietly preserved. Since this app had processed hundreds of thousands of jobs over time, including all error notifications, which also went through the job system, things spiraled. Once the disk was full, new jobs couldn't run, including error reporting. We were blind to the failure.

To make matters worse, recovering the machine wasn't easy. With no space left, even SSH access became unreliable. Freeing up disk space in that state is non-trivial, especially if you don't have automated clean-up in place.

Yes, we should have monitored disk usage more closely. But we did clean up our own records — we just didn’t realize that finished jobs were being preserved automatically.
Proposal

## Should SolidQueue.preserve_finished_jobs really default to true?

Most job libraries don’t retain finished jobs unless you explicitly ask for that. For example, Sidekiq and Resque both discard successful jobs by default. It’s easy to opt into persistence if you need auditing or observability, but it's safer to start with a clean slate.

For a lot of teams, especially those adopting `SolidQueue` for the first time, this default might feel invisible until it causes problems. Would it be worth reconsidering?

stefanvermaas:
> The current default is the right one. Not deleting jobs right after they're executed is more performant (as you don't have to do multiple deletes) and **necessary to ensure recurring tasks aren't enqueued twice**.

I knew there was a good reason for this. This makes sense. 

> The right way to approach this, as I see it, is to periodically clean finished jobs. 

Truthfully, I was actually already aware of this. But I forgot, and then this happened. The reason is that I never thought about it twice because we're still using `redis` + `sidekiq` for most projects. 

> So what I could see is having this recurring task added by default to the generated `recurring.yml` file.

[...]

PR #562, opened 2025-05-13
Automatically clean-up old, finished background jobs

stefanvermaas:
This PR is a suggestion to automatically clean up old, finished background jobs. This prevents people from shooting themselves in the foot as I did. 

See the discussion here: https://github.com/rails/solid_queue/issues/560.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,7 +365,7 @@ There are several settings that control how Solid Queue works that you can set a
 - `silence_polling`: whether to silence Active Record logs emitted when polling for both workers and dispatchers—defaults to `true`.
 - `supervisor_pidfile`: path to a pidfile that the supervisor will create when booting to prevent running more than one supervisor in the same host, or in case you want to use it for a health check. It's `nil` by default.
 - `preserve_finished_jobs`: whether to keep finished jobs in the `solid_queue_jobs` table—defaults to `true`.
-- `clear_finished_jobs_after`: period to keep finished jobs around, in case `preserve_finished_jobs` is true—defaults to 1 day. **Note:** Right now, there's no automatic cleanup of finished jobs. You'd need to do this by periodically invoking `SolidQueue::Job.clear_finished_in_batches`, which can be configured as [a recurring task](#recurring-tasks).
+- `clear_finished_jobs_after`: period to keep finished jobs around, in case `preserve_finished_jobs` is true — defaults to 1 day. **Note:** `SolidQueue` automatically cleans up the finished jobs at 1 AM. To disable this, remove the `automatically_cleanup_finished_jobs` periodic job from the `recurring.yml` configuration.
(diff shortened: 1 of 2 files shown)


lib/generators/solid_queue/install/templates/config/recurring.yml
```diff
@@ -8,3 +8,8 @@
 #     command: "SoftDeletedRecord.due.delete_all"
 #     priority: 2
 #     schedule: at 5am every day
+production:
+  automatically_cleanup_finished_jobs:
+    class: SolidQueue::Job.clear_finished_in_batches
```
calvinchieng, line 13: To use command here.
`command: "SolidQueue::Job.clear_finished_in_batches"`

README.md
```diff
@@ -365,7 +365,7 @@ There are several settings that control how Solid Queue works that you can set a
 - `silence_polling`: whether to silence Active Record logs emitted when polling for both workers and dispatchers—defaults to `true`.
 - `supervisor_pidfile`: path to a pidfile that the supervisor will create when booting to prevent running more than one supervisor in the same host, or in case you want to use it for a health check. It's `nil` by default.
 - `preserve_finished_jobs`: whether to keep finished jobs in the `solid_queue_jobs` table—defaults to `true`.
-- `clear_finished_jobs_after`: period to keep finished jobs around, in case `preserve_finished_jobs` is true—defaults to 1 day. **Note:** Right now, there's no automatic cleanup of finished jobs. You'd need to do this by periodically invoking `SolidQueue::Job.clear_finished_in_batches`, which can be configured as [a recurring task](#recurring-tasks).
+- `clear_finished_jobs_after`: period to keep finished jobs around, in case `preserve_finished_jobs` is true — defaults to 1 day. **Note:** `SolidQueue` automatically cleans up the finished jobs at 1 AM. To disable this, remove the `automatically_cleanup_finished_jobs` periodic job from the `recurring.yml` configuration.
```
elasticspoon, line 368: I think these changes to the README could be a little misleading. Even with these changes SolidQueue won't actually clean up the jobs. Instead when you **install** SolidQueue a job will be created that will clean up jobs. 

I could see someone that ran the generator before this change being confused because their jobs aren't being automatically cleaned up and that code won't be in the `recurring.yml`.

elasticspoon (commented):
I ran into a very similar issue and was thinking about opening a [similar PR](https://github.com/rails/solid_queue/pull/564). :smile:

rosa:
Thanks both! I'm going to go with something intermediate to this, maybe running this every hour, and with a sleep interval as well. I'm worried this might introduce some unexpected load for people by surprise 😅

rosa:
I ended up doing this but with some changes, with a different interval, and different wording in the README, in https://github.com/rails/solid_queue/commit/6437617966bc5bf92995986fc1b14e00f8c68848. Thank you!

PR #565, opened 2025-05-18
feat: update supported ruby versions

elasticspoon:
- Drops support for Ruby 3.1 since it is EOL.
- Tests only first and last versions of 3.2.*
- Tests all current versions of Ruby 3.3.* and 3.4.*

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,17 +24,21 @@ jobs:
       fail-fast: false
       matrix:
         ruby-version:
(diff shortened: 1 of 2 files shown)


rosa:
Hey @elasticspoon, thanks a lot for this! I'm hesitant to drop support for Ruby 3.1 now because that should come with a major version bump, and I want to keep the major version bump for a little bit further in the future, so I'm going to close this one for now. Thank you!

PR #574, opened 2025-06-10
Support Active Job Continuations

djmb:
[Active Job Continuations](https://github.com/rails/rails/pull/55127) will be added in Rails 8.1. Queue adapters need to implement the `stopping?` method to support them.

Implement the method and set it to true when a worker is stopping.

Also add Rails versions so we can test against Rails main. To prevent an explosion of combinations, I've switched to only testing against minor versions of Ruby.


djmb:
The failures against Rails main are due to a Rails bug - see https://github.com/rails/rails/issues/55178.

PR #580, opened 2025-06-17
Add option to skip recurring jobs via environment variable

### LINKED ISSUE #558: Add environment variable to turn off recurring jobs
Hello, thank you for Solid Queue!

We're in the process of migrating an older app form Sidekiq to Solid Queue. We're using Heroku Review apps, and we don't want any recurring jobs to run there. It would be nice if we could completely skip recurring tasks by setting an environment variable.

In searching for a workaround, I found that I could make it work if I added this line to `bin/jobs`:

```diff
  require_relative "../config/environment"
  require "solid_queue/cli"

+ ARGV << "--skip-recurring" if ENV["SOLID_QUEUE_SKIP_RECURRING"]

  SolidQueue::Cli.start(ARGV)
```

(This is of course not the _right_ way to add this, but as a workaround, it's okay.)

Is an option like this something you'd be interested in? I could try to implement it properly, similar to how the `SOLID_QUEUE_RECURRING_SCHEDULE` environment variable works, and send a PR.

Thank you again!

iamradioactive:
Closes #558

**Problem**
Users need a way to disable recurring jobs in specific environments (staging, review apps, development) without modifying configuration files or CLI commands. This is particularly useful for:

- Staging environments where you don't want production cron jobs running
- Review/preview apps that should focus on testing features, not running scheduled tasks
- Development environments where recurring jobs might interfere with testing

**Solution**
- Adds support for the SOLID_QUEUE_SKIP_RECURRING environment variable that mirrors the existing --skip-recurring CLI option functionality.

**Changes**
- **Configuration**: Modified default_options in Configuration class to read SOLID_QUEUE_SKIP_RECURRING environment variable
- **Tests**: Added test coverage with new with_env test helper
- **Documentation**: Updated README with usage examples and use cases

Usage
```
SOLID_QUEUE_SKIP_RECURRING=true bin/jobs
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,8 @@ By default, Solid Queue will try to find your configuration under `config/queue.
 bin/jobs -c config/calendar.yml
 ```
 
+You can also skip all recurring tasks by setting the environment variable `SOLID_QUEUE_SKIP_RECURRING=true`. This is useful for environments like staging, review apps, or development where you don't want any recurring jobs to run. This is equivalent to using the `--skip-recurring` option with `bin/jobs`.
+
 This is what this configuration looks like:
 
 ```yml
@@ -570,6 +572,8 @@ Solid Queue supports defining recurring tasks that run at specific times in the 
 ```
 bin/jobs --recurring_schedule_file=config/schedule.yml
 ```
+
+You can completely disable recurring tasks by setting the environment variable `SOLID_QUEUE_SKIP_RECURRING=true` or by using the `--skip-recurring` option with `bin/jobs`.
 
 The configuration itself looks like this:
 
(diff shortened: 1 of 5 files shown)


iamradioactive:
Checks are failing because of an unrelated single integration spec which is seen failing on `main` branch too. 

Reference: https://github.com/rails/solid_queue/actions/runs/15707068967/job/44255312757

rosa:
> Checks are failing because of an unrelated single integration spec which is seen failing on main branch too.

Yes, it's due to a bug in Rails's main, they only fail with `rails_main`.

rosa (approved):
Thank you!

Issue #582, opened 2025-06-19
After bumping SolidQueue from 1.1.0 to 1.1.1 supervisor process doesn't start


AlexanderNosenko:
We're trying to upgrade gem version from 1.1.0 to 1.1.1 and SolidQueue seems to be silently failing on startup. No processes are launched no failure logs are records. We're running on Ubuntu 22.04.3 x86 with standard unshared MySql 8, rails 7.2.2.1. Not even what details would you need to help us...
Thanks

AlexanderNosenko:
If STDOUT is not collected in your logs, startup failure message would be lost. The instrumentation of Supervisor startup begins after the configuration is validated, so logger doesn't have a chance to catch these.

rosa:
Hey @AlexanderNosenko, in the case of a configuration error, [the process calls `abort`](https://github.com/rails/solid_queue/blob/6437617966bc5bf92995986fc1b14e00f8c68848/lib/solid_queue/supervisor.rb#L18), and the message passed to `abort` is written to `STDERR`. You didn't see anything written there? It's intentional that this is not logged to `STDOUT` 🤔

AlexanderNosenko:
I see, that's why we haven't seen it. We only capture stuff that's logged (via LogSubscriber) and not `STDOUT` or `STDERR`. 
If this is by design, i guess we'll have to capture STD output as well then.

rosa:
Yes, it's by design. Thank you!

PR #583, opened 2025-06-20
Log Supervisor validation errors

AlexanderNosenko:
Issue: https://github.com/rails/solid_queue/issues/582

TLDR: Since v1.1.1 SolidQueue silently (kinda) fails on startup. No processes are launched no failure logs are recorded.

If STDOUT is not collected in your logs, startup failure message would be lost. The instrumentation of Supervisor startup begins after the configuration is validated, so logger doesn't have a chance to catch these.

This is a solution that we think might help some teams, it would definitely have helped us. Please feel free to propose a better solution of course!


rosa:
Hey @AlexanderNosenko, thanks for this! I think, since errors when starting the supervisor go to `STDERR`, I think that's enough, without need for explicit instrumentation there. 

Thank you, nonetheless!

PR #586, opened 2025-06-25
Allow duplicate jobs to be discarded

### LINKED ISSUE #176: Discard duplicate jobs

mhenrixon:
Should resolve #176

I want to prevent the following:

CleanShot 2025-06-25 at 09 26 15@2x

These jobs can under no circumstances end up in this situation, discarding the new jobs instead of queuing them ensures this won't happen. 

I took the simplest possible approach using terminology from sidekiq-unique-jobs. All feedback appreciated.

### DIFF
--- a/app/models/solid_queue/job/concurrency_controls.rb
+++ b/app/models/solid_queue/job/concurrency_controls.rb
@@ -8,7 +8,7 @@ module SolidQueue
       included do
         has_one :blocked_execution
 
-        delegate :concurrency_limit, :concurrency_duration, to: :job_class
+        delegate :concurrency_limit, :concurrency_duration, :concurrency_on_conflict, to: :job_class
 
         before_destroy :unblock_next_blocked_job, if: -> { concurrency_limited? && ready? }
       end
--- a/app/models/solid_queue/job/executable.rb
+++ b/app/models/solid_queue/job/executable.rb
@@ -67,7 +67,12 @@ module SolidQueue
       def dispatch
         if acquire_concurrency_lock then ready
         else
-          block
+          case job_class.concurrency_on_conflict
+          when :discard
+            discard_on_conflict
+          else
+            block
+          end
         end
       end
 
@@ -102,6 +107,10 @@ module SolidQueue
       private
         def ready
           ReadyExecution.create_or_find_by!(job_id: id)
+        end
+
+        def discard_on_conflict
+          finished!
         end
 
         def execution
--- a/lib/active_job/concurrency_controls.rb
+++ b/lib/active_job/concurrency_controls.rb
@@ -12,14 +12,16 @@ module ActiveJob
 
       class_attribute :concurrency_limit
       class_attribute :concurrency_duration, default: SolidQueue.default_concurrency_control_period
+      class_attribute :concurrency_on_conflict, default: :block
     end
 
     class_methods do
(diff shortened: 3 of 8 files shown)


rosa:
Hey @mhenrixon, thanks for this! There was already a PR for it: https://github.com/rails/solid_queue/pull/523
But this idea of marking the job as finished is really clever! It'd allow for quite a few simplifications 🤔  I wonder if it's _too_ clever, though.... would it lead to confusing scenarios where you don't know what happened to a job? Although I guess the same is true if the job is simply discarded, so this should be ok!  

I need to review what happens here for the bulk enqueue case, which was the tricky part of that other PR.

mhenrixon:
> Hey @mhenrixon, thanks for this! There was already a PR for it: https://github.com/rails/solid_queue/pull/523

I need to learn to read all the issues. Sorry about the extra PR.

> But this idea of marking the job as finished is really clever! It'd allow for quite a few simplifications 🤔 I wonder if it's too clever, though.... would it lead to confusing scenarios where you don't know what happened to a job? Although I guess the same is true if the job is simply discarded, so this should be ok!

After working on sidekiq-unique-jobs, and with sidekiq. I have learned that some visibility is appreciated. One of the reasons to use a database-backed queue system is to maintain visibility and avoid paying a blackmailer for keeping the data safe.

I am happy to close this if you prefer the other one @rosa. I didn't look into the other PR in detail, but the bulk enqueue sounds important.

rosa:
Oh, no worries! I think the idea here is better than what we were trying to do in the other PR, so I think combining these two would be the best! I need to review the other one again after some changes so I'll review these two together and will think how to proceed, maybe transplanting some commits over 🤔

mhenrixon:
Feel free to add commits or let me know what to pick or how I can help land this in the main branch. I am at your service!

elias19r:
> would it lead to confusing scenarios where you don't know what happened to a job? Although I guess the same is true if the job is simply discarded, so this should be ok!

I just read this and thought: would it make sense, or be worth it, to add some kind of metadata field to the job to indicate that it has been discarded?

mhenrixon:
> I just read this and thought: would it make sense, or be worth it, to add some kind of metadata field to the job to indicate that it has been discarded?

I thought about that but opted for the simplicity while maintaining visibility. 

I am open to the idea though.

elias19r:
Just throwing out another idea: Alternatively, instead of implementing `concurrency_on_conflict:` with `:block`/`:discard`, what do you think of implementing a `concurrency_block_limit:` config like the following?

- If `concurrency_block_limit: nil`, it blocks new job executions indefinitely (current behavior)
- If `concurrency_block_limit: 0`, it doesn't block new job executions: it discards them right away (this PR's `:discard` behavior)
- If `concurrency_block_limit:` is greater than `0`, it would block executions until the limit of blocked executions is reached, and after that it starts discarding jobs

I didn't check yet if it's possible or worth it. But it looks like we'd need to count the number of blocked execution records after we `acquire_concurrency_lock`

rosa:
Thanks @elias19r for your input! I decided to keep this simple for now. 

I've cherry-picked the changes here and created a new PR: https://github.com/rails/solid_queue/pull/594. Thanks a lot @mhenrixon for your help with this 🙏

PR #592, opened 2025-06-26
More leanient wait time for continuation test

mhenrixon:
This should fix the tests against main @rosa. Curious that the extra wait time is needed but it is what it is.


rosa:
Ohh, interesting! Thanks @mhenrixon, I'll take a look!

rosa:
(I thought the tests failing against `main` were due to https://github.com/rails/rails/issues/55178, but this seems to be something else because that one has been fixed already).

PR #594, opened 2025-07-10
Add support for a `discard` option in concurrency controls

rosa:
This PR builds on #523 and #586, after having great discussions in those with @joelzwarrington and @mhenrixon. Thanks a lot for your work on that and for bearing with me 🙏 In the beginning, I wasn't sure how to approach this because of the way Solid Queue enqueues jobs in two steps (first, the job is created, then an execution for it is created), but after going through those two PRs again and looking into how Sidekiq's unique jobs feature behaves, I think the behaviour here is the right one. 

The way this works is quite simple: a new option, `on_conflict`, that defaults to `:block` but can be set to `:discard`. 

```ruby
class DeliverAnnouncementToContactJob < ApplicationJob
  limits_concurrency to: 1, key: ->(contact) { contact.account }, duration: 5.minutes, on_conflict: :discard

  def perform(contact)
    # ...
```

Basically, when a job needs to be discarded because of a conflict, we delete it in the same transaction where it was created. We don't mark it as finished as #586 does because I don't want to have a provider job ID in this case. I became convinced of this after reviewing how Sidekiq's unique jobs feature behaves. I was super lucky I got the chance to speak with @mperham yesterday at RailsConf (🙏 🙇‍♀️ ), so I could make sure I got this right. Chatting with him and other super nice and helpful attendees (I'm sorry I didn't catch your GitHub handle, but if you're reading this, you know who you are so let me know!), made me realise Active Job's  behaviour with regards to [`successfully_enqueued?`](https://api.rubyonrails.org/v8.0.2/classes/ActiveJob/Core.html#method-i-successfully_enqueued-3F) is inconsistent between `perform_later` and `perform_all_later`. With `perform_later`, unless the adapter raises, the job is considered successfully enqueued. This doesn't match `perform_all_later` where raising doesn't make sense because you need to enqueue multiple jobs, so you don't want the whole thing to fail if a single job fails to be enqueued. In Sidekiq's case, when a job fails to be enqueued an exception is not raised: the return result is not a job ID (`jid`) but `nil` instead. This is reflected in [the bulk enqueue implementation](https://github.com/rails/rails/blob/852190ba265735c0b3b85495020d680312b601cb/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L54-L60) but [not in the individual enqueue](https://github.com/rails/rails/blob/852190ba265735c0b3b85495020d680312b601cb/activejob/lib/active_job/queue_adapters/sidekiq_adapter.rb#L34-L37). I think this should be fixed in Active Job, giving the adapter a chance to set `sucessfully_enqueued` instead of having to raise an exception to prevent it from being set to `true` [here](https://github.com/rails/rails/blob/852190ba265735c0b3b85495020d680312b601cb/activejob/lib/active_job/enqueuing.rb#L134-L139). I think the inconsistency might simply be because `perform_all_later` was added much later and the exception raising behaviour couldn't be used there 🤔 

Anyway, here I chose to do the same as Sidekiq, and just not enqueue the job at all (and not have a job ID) when we need to discard a job because of concurrency limits.


joelzwarrington:
Awesome! 👏

In the future, I also see an option to: `on_duplicate: :terminate_existing`

mhenrixon:
It looks great, @rosa! Thank you for wrapping it up for us; I am happy to have contributed some ideas. 

Look forward to seeing this released. ❤️

Teyler7:
This looks incredible! 

It was a honor to get to talk to you and Mike at Rails Conf and discuss Active Job's `successfully_enqueued?` behavior. Thanks for all your hard work on Solid Queue 🧡

app/models/solid_queue/job/concurrency_controls.rb
```diff
@@ -46,6 +50,14 @@ def release_concurrency_lock
           Semaphore.signal(self)
         end
 
+        def handle_concurrency_conflict
+          if concurrency_on_conflict.discard?
+            destroy
```
chaadow, line 55: `destroy!` maybe?

PR #599, opened 2025-07-18
Fix the devcontainer configuration

rafaelfranca:
The name of the config file was wrong, and the tests were not possible to run in this project inside that container.


rosa:
Ohh, thank you @rafaelfranca! 🙏

PR #600, opened 2025-07-18
Don't lock thor version to 1.3.x

rafaelfranca:
This is blocking me to upgrade Rails to use thor 1.4.0.

None of the Rails dependencies have upper bounds on any gem to avoid this kind of issues.


blakegearin:
I believe this needs to be released in order to resolve [CVE-2025-54314](https://github.com/advisories/GHSA-mqcp-p2hv-vw6x)

PR #601, opened 2025-07-21
Link `claimed_executions` to `processes` via `process_name`

rosa:
That is, instead of linking them via the standard `process_id` (`claimed_execution.process_id` and `process.id`). 

The reason for this change is that we're exploring the use of Solid Queue in a multi-tenanted setup, where we'll have an individual database per tenant, but a shared supervisor, dispatcher, scheduler, and workers for multiple tenants. We'll keep a supervisor-local database separated from the tenanted DBs, with just the `solid_queue_processes` table. The supervisor will control the processes stored in that table in that local DB, whereas each tenant will use its own database with the remaining tables, which will contain the jobs and the job metadata.

The consequence of this is that we can no longer rely on finding a process's in-flight jobs via its `id`, because the `id` can be (and most likely will be) duplicated across supervisors' local databases. The name, however, is randomly generated via `SecureRandom.hex(10)`, which, with a very high likelihood, guarantees unique names across supervisors' local DBs. In this way, if a process is killed, leaving in-flight jobs (claimed executions) behind, and we fail over a different supervisor with a different local DB, that supervisor will be able to know which claimed executions are orphaned. 

cc @flavorjones


lib/generators/solid_queue/update/templates/db/link_claimed_executions_with_processes_through_name.rb
```diff
@@ -0,0 +1,30 @@
+class LinkClaimedExecutionsWithProcessesThroughName < ActiveRecord::Migration[<%= ActiveRecord::VERSION::STRING.to_f %>]
+  def up
+    unless connection.column_exists?(:solid_queue_claimed_executions, :process_name)
+      add_column :solid_queue_claimed_executions, :process_name, :string
+      add_index :solid_queue_claimed_executions, :process_name
+    end
```
rosa, line 6: This migration is idempotent because it is already included in the setup schema, for people who are installing Solid Queue for the first time. If there's a new update in the future with a new migration, people should be able to run all of them without any issues.

rosa:
Looks like we won't be needing this in the end! I might cherry-pick some bits that could come handy in the future to introduce new migrations, for example for batches.

flavorjones:
@rosa Totally good to close this. I hope we will want something like it eventually (I still want to get the concept of active/passive upstreamed into Rails) but we can always resurrect this branch if/when that happens. ❤️

rosa:
Thank you @flavorjones! Yes, absolutely, happy to resurrect this in the future!

PR #610, opened 2025-07-23
Remove Rakefile from gemspec files list

smasato:
This PR removes `Rakefile` from the list of files included in the gem package. I think that `Rakefile` is only needed for development and testing purposes and does not need to be distributed with the gem.


rosa:
Oh yes! You're totally correct, I think that was from the engine generator, I never removed it 😅

rosa:
Thank you!

PR #611, opened 2025-07-23
Supress sqlite fork warnings for rails 7.1 & 7.2

p-schlickmann:
When sqlite3 > 2.1.0 and rails < 8.0.0 we have the following warning when using Solid Queue:

```
.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/sqlite3-2.5.0-x86_64-linux-gnu/lib/sqlite3/fork_safety.rb:43: warning: Writable sqlite database connection(s) were inherited from a forked process. This is unsafe and the connectio
ns are being closed to prevent possible data corruption. Please close writable sqlite database connections before forking.
```

This is polluting our CI logs and triggering warnings for people using Solid Queue with Rails 7.1 and 7.2 - although the behaviour is safe: https://github.com/rails/solid_queue/issues/324#issuecomment-2372245070

Also mentioned in: https://github.com/rails/solid_queue/issues/506

### DIFF
--- a/test/dummy/config/initializers/sqlite3.rb
+++ b/test/dummy/config/initializers/sqlite3.rb
@@ -27,4 +27,8 @@ ActiveSupport.on_load :active_record do
     ActiveRecord::ConnectionAdapters::SQLite3Adapter.prepend SqliteImmediateTransactions
     ActiveRecord::ConnectionAdapters::SQLite3Adapter.prepend SQLite3Configuration
   end
+
+  # Suppress fork safety warnings in the dummy app (used by test suite)
+  # Warnings are safe: https://github.com/rails/solid_queue/issues/506
+  SQLite3::ForkSafety.suppress_warnings! if defined?(SQLite3::ForkSafety)
 end


p-schlickmann:
Actually, the warning is still showing up in the CI - wait a sec.

p-schlickmann:
Done! No more warnings in the CI as well: https://github.com/rails/solid_queue/actions/runs/16473942583/job/46570298080?pr=611

rosa:
Hey @p-schlickmann, thanks a lot for this! I think I'd keep only the call for tests only, in `test/test_helpers/processes_test_helper.rb`, but not the one in the supervisor code. I'd leave that to Solid Queue users on Rails < 8 to decide whether they want to call that or not (and maybe they can do it in the `on_start` hook, because I think the warnings happen on fork, and `on_start` runs before forking in the Supervisor 🤔 ).

p-schlickmann:
Yes, @rosa, that approach should work! 

But in addition to cleaning up the CI logs, the goal of this PR is to reassure Solid Queue users on Rails 7.1 and 7.2 that the forking warning is harmless and can be safely ignored.

Maybe it makes sense to open a PR against the [Rails](https://github.com/rails/rails) repo to suppress this warning upstream in Rails 7 itself? Like they did for Rails 8?

rosa:
Maybe, yes, I'd prefer that if the core team allows it.

p-schlickmann:
@rosa, unfortunately we can’t apply the same fix in Rails 7 because `SQLite3::ForkSafety.suppress_warnings!` was included in `sqlite3 >= 2.1`, while Rails 7 minimum supported version is `1.4`. 

I don't think a warning is gonna make the rails team bump such a major version like that.

In any case, I’ve pushed a commit to suppress those warnings in our CI, so our test output is now completely clean 🎉

rosa (approved):
Awesome, thank you!

Issue #602, opened 2025-07-21
Flaky tests


p-schlickmann:
Hi team,

I understand the current state of the test suite is not ideal - some tests are flaky.

I'd love to help improve this. Ensuring test reliability should be a top priority, especially for an open-source tool that's the default queue mechanism in Rails 8.

@rosa I created this issue so we can discuss it further - do you have any advice or suggestions before I start digging in?

Thanks!

rosa:
Hey @p-schlickmann, yes. All these tests that fork processes and rely on waiting on things to happen elsewhere are quite flaky indeed. I don't have any particular suggestions to start looking, I'm afraid, but any help here would be quite welcome!

rosa:
Looks like the flakiest test after the last improvements is this one: 
```
bin/rails test test/integration/instrumentation_test.rb:306
```
I tried to make it better in https://github.com/rails/solid_queue/commit/f433a99fd402dd25213db445dbcebef29617aa8c but that didn't work so well. I'll look into it later.

p-schlickmann:
Yea I noticed `test/integration/processes_lifecycle_test.rb` is also quite flaky - I'm working on it

diogovernier:
Hey folks! 👋 

I opened this PR to address a flaky test we've been seeing: https://github.com/rails/solid_queue/pull/733

It's related to this failure: https://github.com/rails/solid_queue/actions/runs/24190892956/job/70607570480?pr=732

```
Run bin/rails test
Run options: --seed 8690

# Running:

bundler: failed to load command: puma (/home/runner/work/solid_queue/solid_queue/vendor/bundle/ruby/3.1.0/bin/puma)
/home/runner/work/solid_queue/solid_queue/vendor/bundle/ruby/3.1.0/gems/puma-7.2.0/lib/puma/binder.rb:344:in `initialize': Address already in use - bind(2) for "127.0.0.1" port 9222 (Errno::EADDRINUSE)
```

Would love your thoughts on the approach. Happy to explore a different direction if this doesn't feel right.

Also noticed another flaky test failing in the PR (seems unrelated to the Puma issue). I'm happy to dig into that as well.

rosa:
[...]

PR #619, opened 2025-08-06
attempt to reduce flakiness of `processes_lifecycle_test.rb`

p-schlickmann:
Hey @rosa 👋 

This is an attempt to reduce the flakiness of `processes_lifecycle_test.rb`

### DIFF
--- a/test/integration/processes_lifecycle_test.rb
+++ b/test/integration/processes_lifecycle_test.rb
@@ -127,10 +127,8 @@ class ProcessesLifecycleTest < ActiveSupport::TestCase
 
     signal_process(@pid, :TERM, wait: 0.5)
 
-    sleep(SolidQueue.shutdown_timeout + 0.5.second)
-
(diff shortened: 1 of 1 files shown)


p-schlickmann:
Well, that clearly didn't work lol

This is quite tricky! I ran it over 5k times locally, no flakiness.

I'm starting to think about alternative approaches, maybe we can use [minitest-retry](https://github.com/y-yagi/minitest-retry) gem?

rosa:
Sorry @p-schlickmann, I missed this one! Let's try a bit more to see if we can solve the ones pending, and then if not, we can try with minitest-retry. I'm worried that gem will, long-term, hide all the flaky tests and turn the whole test suite into something pretty slow 😅 (It has happened to me in the past). 

Going to close this one, since this attempt didn't work. I'll keep working on this myself from time to time!

p-schlickmann:
No worries @rosa, been quite busy here as well. 

Yes, I do agree we can try a bit more until adding `minitest-retry`. But if we do, maybe we can set max retries to a very small number so that problem you mentioned doesn't happen in the long run.

Issue #528, opened 2025-03-06
Port 33060 is used by MySQL's X Protocol


brendon:
You'll need to choose another port for your dev setup as it conflicts with a locally installed MySQL if the X Protocol port is running.

33061 would be find but I'll let you pick :D

ghost:
@rosa do you guys want this to be changed? If so, I can try to do it

Also, in case this change is needed, do you have any suggestions on how can/should I verify it, so I can check it didn't break anything?

brendon:
It shouldn't break anything I don't think. Just needs the configuration changed in the test suite and the Dockerfile:

https://github.com/rails/solid_queue/blob/6dc5c6681fecb0b3c712bf24d89ddbbc43a4b14b/test/dummy/config/database.yml#L36

ghost:
@brendon, yes I understand that :)

Wanted Rosa'a opinion on this one.

PR #626, opened 2025-08-19
fix: Change MySQL port to 33066

ivnhk:
Resolves: https://github.com/rails/solid_queue/issues/528

In order to avoid conflict with MySQL Protocol X, the port was changed to `33066`.

The change was tested locally by running `docker compose up`


ghost:
I am not sure why those 2 particular tests failed. Seems like the issue with dispatchers.

rosa (approved):
Thank you! yeah, the tests are flaky, nothing to do with this change. We're improving the flaky tests but aren't finished yet 🙏

PR #634, opened 2025-09-04
Correctly handle jobs when ActiveSupport.parse_json_times is enabled

npezza93:
Reimplementing ActiveRecord::Coders::JSON is definitely not my favorite fix so im open to any ideas on how to clean this up.

But if you set `ActiveSupport.parse_json_times = true` an error gets thrown when running jobs because of this line in active_job/core.rb:170:
```
      self.enqueued_at          = Time.iso8601(job_data["enqueued_at"]) if job_data["enqueued_at"]
```

When parse_json_times is enabled job_data["enqueued_at"] is an AS::TimeWithZone object instead of a string.
Alternatively, I could send a patch to Rails to force it to be a string: `Time.iso8601(job_data["enqueued_at"].to_s)` but thought I would start here and get thoughts since that fix could impact all ActiveJob adapters.


npezza93:
Thinking about this more, I think this should be handled upstream in ActiveJob. Going to try fixing it there instead.

PR #652, opened 2025-10-08
Replace use of `ActiveRecord::Base.connection` with `with_connection`

### LINKED ISSUE #646: Avoid using bare connection.
When running with:

```ruby
config.active_record.permanent_connection_checkout = :disallowed
```

The following code appears to have issues:

https://github.com/rails/solid_queue/blob/1957f570ad4bce146516435ba1dfe5d270b8363e/app/models/solid_queue/recurring_task.rb#L39

Better to use `Model.with_connection do |connection|`.

See <https://github.com/socketry/community/discussions/8> for background.

rosa:
So that when running with `permanent_connection_checkout` set to `false`, this doesn't cause any problems. Using `connection` is deprecated in any case, so this was needed regardless.

Fixes #646

### DIFF
--- a/app/models/solid_queue/failed_execution.rb
+++ b/app/models/solid_queue/failed_execution.rb
@@ -58,8 +58,11 @@ module SolidQueue
       end
 
       def determine_backtrace_size_limit
-        column = self.class.connection.schema_cache.columns_hash(self.class.table_name)["error"]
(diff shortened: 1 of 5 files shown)


ioquatix:
Awesome, thanks!

PR #654, opened 2025-10-20
Second attempt reduce flakiness of `processes_lifecycle_test.rb`

p-schlickmann:
Second attempt to reduce the flakiness of  `processes_lifecycle_test.rb`.

I started using the same logic from [#L65-L85](https://github.com/rails/solid_queue/blob/main/test/integration/processes_lifecycle_test.rb#L65-L85
).

According to the comment, we should actually check for `:ready` rather than `:claimed`.

Issue https://github.com/rails/solid_queue/issues/602


p-schlickmann:
@rosa Seems to be working now! The failed test is unrelated to `processes_lifecycle_test.rb` 🙏🏻

rosa (approved):
Nice! Thank you!

PR #655, opened 2025-10-20
Wrap` Supervisor#start` and `stop` with the app executor

flavorjones:
Wrap` Supervisor#start` and `stop` with the app executor to support applications that are using `Executor` callbacks or setting a custom app executor.


rosa:
Going to revert this one because it caused problems in dev mode (see the linked issue), while we figure out the best way to fix it (using the reloader or doing something different).

PR #661, opened 2025-10-22
print warning on startup if path to configuration file does not exist

### LINKED ISSUE #632: Raise an error if config file path does not exist?
Hi there 👋

Today I learned that if `-c` resp. `--config-file` option points to a path that doesn't exist, it falls back to a default config.

```sh
bin/jobs --config-file="config/this-file-does-not-exist.yml"
```

I wonder if it should rather raise an error if the file does not exist?
Or maybe a warning that it falls back to a default?

lxxxvi:
Hi 👋 

Here's my attempt to resolve #632 

I suggest to display a WARNING in the terminal if the path to configuration file does not exist:

```text
$ bin/jobs -c /path/to/nowhere.yml
[solid_queue] WARNING: Provided configuration file '/path/to/nowhere.yml' does not exist. Falling back to default configuration.

```

Let's discuss the approach and, of course, the message itself.

Looking forward to hearing your feedback.

### DIFF
--- a/test/unit/configuration_test.rb
+++ b/test/unit/configuration_test.rb
@@ -29,7 +29,7 @@ class ConfigurationTest < ActiveSupport::TestCase
   test "warns if provided configuration file does not exist" do
     assert_output "[solid_queue] WARNING: Provided configuration file '/path/to/nowhere.yml' does not exist. Falling back to default configuration.\n" do
       configuration = SolidQueue::Configuration.new(config_file: Pathname.new("/path/to/nowhere.yml"))
-      assert_predicate configuration, :valid?
+      assert configuration.valid?
     end
   end
 


test/unit/configuration_test.rb
```diff
@@ -26,6 +26,13 @@ class ConfigurationTest < ActiveSupport::TestCase
     assert_processes configuration, :dispatcher, 1, batch_size: SolidQueue::Configuration::DISPATCHER_DEFAULTS[:batch_size]
   end
 
+  test "warns if provided configuration file does not exist" do
+    assert_output "[solid_queue] WARNING: Provided configuration file '/path/to/nowhere.yml' does not exist. Falling back to default configuration.\n" do
+      configuration = SolidQueue::Configuration.new(config_file: Pathname.new("/path/to/nowhere.yml"))
+      assert_predicate configuration, :valid?
```
rosa: Maybe 
```suggestion
      assert configuration.valid?
```
so it matches other usages of this in this test?

PR #667, opened 2025-10-26
Fix broken `rails_main` tests

faraquet:
On the latest Rails, the tests are broken.
```
Failure:
DispatcherTest#test_polling_queries_are_logged [test/unit/dispatcher_test.rb:48]:
Expected /SELECT .* FROM .solid_queue_scheduled_executions. WHERE/ to match "".

bin/rails test test/unit/dispatcher_test.rb:39

Failure:
WorkerTest#test_polling_queries_are_logged [test/unit/worker_test.rb:114]:
Expected /SELECT .* FROM .solid_queue_ready_executions. WHERE .solid_queue_ready_executions...queue_name./ to match "".

bin/rails test test/unit/worker_test.rb:105
```

### Reason

Rails migrated framework log subscribers to structured events via ActiveSupport::EventReporter, making SQL logs debug-only structured events. See PR: Make framework log subscribers consume structured events (https://github.com/rails/rails/pull/55900/files).

### What broke

Tests that captured SQL via a StringIO logger and matched SELECT statements no longer saw logs unless EventReporter debug mode was enabled and the subscriber logger was pointed at the StringIO.

### Solution
Added `LoggingTestHelper` with:
  - `with_polling` to toggle `SolidQueue.silence_polling` for deterministic polling logs.
  - `with_active_record_logger` to:
    - Always redirect `ActiveRecord::Base.logger` to the test `StringIO`.
    - If Rails uses structured events (≥ 8.1), enable `ActiveSupport.event_reporter.debug_mode` and redirect `ActiveSupport::LogSubscriber.logger` to the `StringIO` using one-line flip: `old_as_ls_logger, ActiveSupport::LogSubscriber.logger = ActiveSupport::LogSubscriber.logger, logger`, restoring afterward.
- Pre-8.1 Rails: falls back to previous behavior (no EventReporter calls).


p-schlickmann:
Great work @faraquet!

But I wonder if Rails is gonna fix this behaviour upstream? Since [other people](https://github.com/rails/rails/pull/55900#issuecomment-3445390380) are also having issues?

Also, the failures are against `rails_main`, which is the unreleased version of `8.2`, not `8.1`.

But yes, I agree we should add `8.1` to the matrix since it was released past Wednesday. Maybe in a separate PR?

faraquet:
Thank you so much @p-schlickmann 

> But yes, I agree we should add 8.1 to the matrix since it was released past Wednesday. Maybe in a separate PR?

Extracted into https://github.com/rails/solid_queue/pull/668

faraquet:
Thanks for the pointer @p-schlickmann! I'll check out today's Rails `main` branch changes and decide based on this. Potentially, it might be resolved on their end

rosa:
Thanks @faraquet! Going with https://github.com/rails/solid_queue/pull/674 in the end.

Issue #670, opened 2025-10-28
Dev mode doesn’t reload changes properly — workers stop or ignore updated code


michelson:
Hi,
I'm not sure how to provide a fully reproducible test case, but I’m seeing some erratic behavior with Solid Queue in development mode.

In my app, I have several service classes under app/services. When I make code changes there, the workers don’t pick them up — in fact, sometimes the queue processing stops entirely. I have to restart the process each time for the changes to take effect.

As a test, I temporarily switched to Sidekiq, and in that setup code changes are picked up right away, so this seems to be related to how Solid Queue handles app reloading in development.

Environment:

Rails version:  8.0.1

Ruby version: 3.5+

Solid Queue version: 1.2.2

Expected behavior:
Workers should reload the updated code automatically in development, or at least behave consistently with Rails reloader.

Actual behavior:
Changes in service files are ignored until I restart the worker process.

rosa:
Hey @michelson, I think this was introduced with the last release. Could you try to see if this works correctly in version 1.2.1? I suspect https://github.com/rails/solid_queue/pull/655 might be the cause 🤔

zarqman:
Seeing the same here on v1.2.2, but with Ruby 3.4 and Rails 8.1. Based on the behavior, I also wondered if it's related to the reloader and #655 caught my eye as well. Currently trying v1.2.1..

Looking at #655, since `Supervisor#start` runs perpetually, it seems like a questionable idea to enter the app_executor and just stay inside for the duration of the process. I don't believe the standard Rails app_executor is intended to be used that way. 

I'm not quite sure of #655's exact intended use case, but perhaps `Supervisor#start` should only wrap `run_start_hooks` (and maybe `boot`), and then `#stop` should only wrap `run_stop_hooks` ?

michelson:
Hola Rosa, thank you for your quick reply. I can confirm that downgrading to version 1.2.1 resolved the code reload issue in my development environment.

rosa:
@michelson, thanks for confirming. I just reverted https://github.com/rails/solid_queue/pull/655, and I'm going to release version 1.2.3 shortly with that change.

rosa:
Relased that, so I'm going to close this one. Thanks for the report!

zarqman:
Wanted to confirm that v1.2.3., with the reversion in #672, is working well again here. Thanks!

PR #673, opened 2025-10-29
Add wrap_in_app_executor in a few necessary places

flavorjones:
`Process#register` and `#degister` for the supervisor to start up properly.

`RecurringSchedule#reload_tasks` resolves all the records immediately to avoid deferred resolution outside the executor block.

`Supervisor#handle_claimed_jobs_by` wraps its code in the executor.

A second attempt at #655 without causing the issues from #670

This PR unfortunately doesn't add any tests; I tried and failed to come up with a strategy that would check all database code was wrapped properly that didn't involve massive restructuring of tests. If you have any good ideas on how to do that, please let me know.

(As an aside, the reason for this PR and #655 is that I'm working on multi-tenant support for Solid Queue, using Active Record Tenanted which relies on Rails' sharding features. I'd like to use the app executor's run/complete callbacks to set/unset the shard.)

### DIFF
--- a/lib/solid_queue/processes/registrable.rb
+++ b/lib/solid_queue/processes/registrable.rb
@@ -18,17 +18,19 @@ module SolidQueue::Processes
       attr_accessor :process
 
       def register
-        @process = SolidQueue::Process.register \
-          kind: kind,
-          name: name,
-          pid: pid,
-          hostname: hostname,
-          supervisor: try(:supervisor),
-          metadata: metadata.compact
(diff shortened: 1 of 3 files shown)


flavorjones:
Some of these tests are failing against Rails main because of the problem fixed in https://github.com/rails/solid_queue/pull/674

The others seem to have been flaky docker behavior.

rosa (approved):
Fantastic, this is great. Thanks a lot @flavorjones!

PR #674, opened 2025-10-29
Fix testing against Rails main

flavorjones:
Default logging levels changed in rails/rails#55900

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -53,6 +53,17 @@ class ActiveSupport::TestCase
     end
   end
 
+  def run(...)
+    # Rails 8.1.dev changed default logging levels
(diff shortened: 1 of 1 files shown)


flavorjones:
I cannot reproduce the test failure locally. It might be a flake? I'm honestly not sure.

rosa:
> I cannot reproduce the test failure locally. It might be a flake? I'm honestly not sure.

Yes! This is one of the flaky tests we've been working lately to address.

rosa (approved):
Ohh, nice! I had this pending but hadn't gotten to it yet. I think this fix makes more sense than https://github.com/rails/solid_queue/pull/667, so I'll close that one.

PR #678, opened 2025-11-03
Minimizing flakiness of `dispatcher_test.rb`

p-schlickmann:
Issue: https://github.com/rails/solid_queue/issues/602  
Flake reference: https://github.com/rails/solid_queue/blob/main/test/unit/dispatcher_test.rb  
Failed run: https://github.com/rails/solid_queue/actions/runs/18917303921/job/54003968729  

### Summary
1. **`polling queries are logged`** — Rewind `StringIO` and increase sleep  
2. **`polling queries can be silenced`** — Rewind `StringIO` and increase sleep  
3. **`silencing polling queries when there's no Active Record logger`** — Increase sleep  
4. **`run more than one instance of the dispatcher`** — Increase `AddToBufferJob` wait and increase sleep  
5. **`sleeps 0.seconds between polls if there are ready to dispatch jobs`** — Increase `AddToBufferJob` wait and increase sleep  

While the CI logs helped diagnose flakes in tests **1** and **2**, the following command was useful to reproduce others locally:

```bash
export TARGET_DB=sqlite
export BUNDLE_GEMFILE=./gemfiles/rails_7_1.gemfile # or whatever
export RAILS_ENV=test
# Run 4000 tests in parallel using 4 CPUs (you must install `parallel` CLI)
seq 1 4000 | parallel -j4 --halt now,fail=1 'TEST_ENV_NUMBER={%} bin/rails test test/unit/dispatcher_test.rb'
```
You might also need to add `TEST_ENV_NUMBER` to `database.yml`:
```yml
test:
  primary:
    <<: *default
    database: <%= database_name_from("test" + ENV["TEST_ENV_NUMBER"]) %>
  shard_one:
    <<: *default
    database: <%= database_name_from("test_shard_one" + ENV["TEST_ENV_NUMBER"]) %>
    migrations_paths: db/migrate_shards
  shard_two:
    <<: *default
    database: <%= database_name_from("test_shard_two" + ENV["TEST_ENV_NUMBER"]) %>
    migrations_paths: db/migrate_shards
  queue:
    <<: *default
    database: <%= database_name_from("test_queue" + ENV["TEST_ENV_NUMBER"]) %>
    migrations_paths: db/queue_migrate
```
And also modify the log path for `environments/test.rb`:
```ruby
Rails.application.configure do
    # ...
    config.paths["log"] = "log/test#{ENV['TEST_ENV_NUMBER']}.log"
end
```

PR #681, opened 2025-11-10
More flakes and prevent deadlocked actions from consuming GH quota

p-schlickmann:
Flake: [processes_lifecycle_test.rb  ](https://github.com/rails/solid_queue/blob/main/test/integration/processes_lifecycle_test.rb)
Instructions to validate the fix locally: https://github.com/rails/solid_queue/pull/678  

### Summary
This PR mainly increases timeout gaps between a few assertions and uses helper methods like `process_exists` and `wait_for_jobs_to_finish_for` to fully eliminate flakiness from `processes_lifecycle_test.rb`.  
This is probably the 3rd or 4th attempt 😆, but I’m confident it's the last one!

It also adds `timeout-minutes: 15` to prevent deadlocked actions from consuming GitHub Actions quota.  
See: https://github.com/rails/solid_queue/actions/runs/18691886640/job/53299407575 (6h duration!)

### DIFF
--- a/test/integration/processes_lifecycle_test.rb
+++ b/test/integration/processes_lifecycle_test.rb
@@ -58,7 +58,7 @@ class ProcessesLifecycleTest < ActiveSupport::TestCase
       signal_process(@pid, :TERM, wait: 0.1.second)
     end
 
-    sleep(1.second)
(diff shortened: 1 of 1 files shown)


rosa (approved):
Brilliant! And great catch on the timeout for GitHub actions 😅

PR #686, opened 2025-11-18
Attempt to reduce flakiness of `scheduler_test`

p-schlickmann:

Attempting to fix the latest [flake against main](https://github.com/rails/solid_queue/actions/runs/19438921829/job/55617036437)


rosa:
Very nice! Sorry for the delay @p-schlickmann, and thanks so much for your continued work on this 🙇🏻‍♀️

PR #687, opened 2025-11-25
Fixed indentation for the adapter line in the README’s cable.yml example

simaaaji:
Hi, while configuring cable.yml for the development environment using the README, I noticed that the indentation in the README would cause an error if used as-is.
It’s a small fix, but to prevent confusion for future readers, I’m sending a pull request.


rosa (approved):
Thank you!

Issue #456, opened 2024-12-16
Job with concurrency key blocked when it shouldn't have, eventually was run after `duration` expired


leondmello:
We are observing that our jobs are sometimes unnecessarily blocked.
In the following example, for the same concurrency key, there are only 2 jobs. The first one completes immediately in just over a 100 milliseconds (`finished_at` value).
The second one takes over 30 minutes, we are guessing it is blocked by the concurrency control duration. The `scheduled_at` time of the second job is much after when the first job completes.
As per our understanding, the second job should have started immediately.
We confirmed that none of the jobs processes bounced during that time.

`...` indicates redacted data.

### Concurrency Config
```ruby
limits_concurrency to: 1, key: ->(action, params) do
  ...
end, duration: 30.minutes
```

### Queue Config
```yaml
dispatchers:
    - polling_interval: 1
      batch_size: 500
workers:
    - queues: [<queue of blocked job>, "*"]
      threads: 3
      processes: <%= ENV.fetch("JOB_CONCURRENCY", 1) %>
      polling_interval: 0.1
```

### Console queries
```ruby
SolidQueue::Job.where("DATE_PART('minutes', finished_at - scheduled_at) >= 30").count
=> 1

SolidQueue::Job.where(concurrency_key: '...').select(:active_job_id, :scheduled_at, :finished_at).order(:finished_at)
=>
[#<SolidQueue::Job:0x00007f062d151a50
  active_job_id: "60916418-e7a3-4f1e-9171-357b6e52c154",
  scheduled_at: "2024-12-16 01:20:09.441199000 +0000",
  finished_at: "2024-12-16 01:20:09.648985000 +0000",
  id: nil>,
 #<SolidQueue::Job:0x00007f062d151910
  active_job_id: "3c4f34da-9db7-430e-9cbe-c1214f98ccb9",
  scheduled_at: "2024-12-16 01:54:49.287841000 +0000",
  finished_at: "2024-12-16 02:24:50.264702000 +0000",
  id: nil>]

 ````
 
 ### Logs about the second job execution
 ```
I, [2024-12-16T02:24:50.154361 #84]  INFO -- : [ActiveJob] [...] [3c4f34da-9db7-430e-9cbe-c1214f98ccb9] Performing ... (Job ID: 3c4f34da-9db7-430e-9cbe-c1214f98ccb9) from SolidQueue(...) enqueued at 2024-12-16T01:54:49.287885003Z with arguments: ...
[...]

PR #689, opened 2025-11-29
Fix duplicate job execution despite limits_concurrency

mhenrixon:
## Summary

Fixes multiple race conditions that allow duplicate concurrent job execution despite `limits_concurrency`. These were discovered while debugging very long-running jobs crashing our production system.

### 1. Blocked jobs released while job is still executing

**Root cause**: When a job exceeds `concurrency_duration`, the semaphore expires and gets deleted by concurrency maintenance. `BlockedExecution.releasable` considered a key releasable when its semaphore was missing — causing blocked jobs to run concurrently with the still-executing job.

**Fix**: Check for claimed executions before marking a key as releasable. A key is only releasable when its semaphore is missing or has available slots AND no jobs with that key are currently executing.

### 2. Race condition between job enqueue and concurrency unblock

**Root cause**: Without a `FOR UPDATE` lock in `Semaphore#wait`, this interleaving caused permanently stuck jobs:
1. Job B enqueue reads semaphore (value=0, no lock) → decides to block
2. Job A finishes: signal → UPDATE value to 1
3. Job A: `release_one` → finds nothing (B's BlockedExecution not committed yet)
4. Job B commit → BlockedExecution exists but nobody unblocks it

**Fix**: `Semaphore.lock.find_by(key: key)` acquires a `FOR UPDATE` lock, serializing the enqueue path with the signal path. The signal UPDATE must wait for the enqueue transaction to commit, guaranteeing the BlockedExecution is visible to `release_one`.

### 3. Concurrency bypass during graceful shutdown

**Root cause**: `ClaimedExecution#release` unconditionally called `dispatch_bypassing_concurrency_limits`, ignoring whether another job with the same key was already running. After semaphore expiry, the released job could run concurrently with an existing execution.

**Fix**: `other_executions_holding_concurrency_lock?` checks for other claimed executions with the same key. If found, dispatch goes through normal concurrency controls. If not, bypass is safe.

### 4. `signal_all` could push semaphore above limit

**Root cause**: `Proxy.signal_all` did `UPDATE SET value = value + 1` without the `value < limit` guard that individual `signal` has. If called on a job whose slot was already returned, the semaphore value could exceed the limit, allowing over-admission.

**Fix**: Group jobs by their `concurrency_limit` and apply the same `WHERE value < limit` guard used by `attempt_increment`.

### 5. Process crash between finish and unblock loses signal

**Root cause**: `finished` and `failed_with` committed their transactions (destroying claimed_execution) separately from `unblock_next_job` (signaling semaphore + releasing blocked job). A process crash between them lost the signal, causing blocked jobs to wait until semaphore expiry (`concurrency_duration`).

**Fix**: Move `unblock_next_job` inside the same transaction as `finished` and `failed_with`. If the process crashes, the whole transaction rolls back — the claimed_execution survives, and the supervisor's process pruning will properly signal via `fail_all_with`.

### 6. Semaphore under-count during shutdown release

**Root cause**: When releasing a claimed execution while other jobs with the same key are executing, the code re-dispatched without first returning the held semaphore slot. This caused the semaphore to permanently under-count available slots, reducing effective parallelism until expiry.

**Fix**: `Semaphore.signal(job)` before re-dispatch returns the slot to the pool. The subsequent `dispatch` re-acquires or blocks as appropriate.

## Test Plan

- [x] All 211 existing tests pass on PostgreSQL (0 failures, 0 errors, 0 skips)
- [x] New test: `signal_all` does not increment semaphore beyond limit
- [x] New test: `signal_all` increments when below limit
- [x] New test: `perform` unblocks next job atomically with finish
- [x] New test: `failed_with` unblocks next job atomically
- [x] New test: semaphore row lock blocks concurrent signal (PostgreSQL)
- [x] New test: blocked execution visible to `release_one` after signal
- [x] New test: don't unblock jobs when semaphore expires but job still executing
- [x] New test: release bypasses concurrency limits when no other job executing
- [x] Integration tests with forked supervisor (3 workers, dispatcher with concurrency maintenance)


jpcamara:
Is the issue that it violates the expectation? Because this seems like expected behavior - if the duration elapsed, then you don't lock the job anymore. That job just exists outside of the concurrency limits now?

mhenrixon:
> Is the issue that it violates the expectation? Because this seems like expected behavior - if the duration elapsed, then you don't lock the job anymore. That job just exists outside of the concurrency limits now?

Exactly, even with a 12 hour limit on the concurrency I saw endless duplicate jobs for two different reasons, both fixed here and one being exactly what you just insinuated. 

Retry, restarts, OOM errors, etc cause a number of duplicates which this PR resolves. 

Ultimately, that worker will be rewritten anyway but it seemed like the locking mechanism has some flaws/bleeds when I looked at it.

rosa:
Hey @mhenrixon, sorry for the delay! I'm just catching up on PRs and issues this week. 

I'm also sorry that I don't follow the issue here. What you describe as flawed is exactly how the system should behave. From the README: 

> jobs will be blocked from running, and they'll stay blocked until another job finishes and unblocks them, or after the set expiry time (concurrency limit's duration) elapses.

That is, if the expiry time finishes, the lock is no longer valid, and other jobs are allowed to run concurrently. This is what `duration` is for. It's a safety mechanism in case you have a job that doesn't finish for some reason, or takes much longer than it should, so that it doesn't block your other jobs forever.

mhenrixon:
I get that but this isn't what I see in the running app where duplicates still show up. Maybe because of restarts, maybe because of other reasons but within that expiry I still see duplicates without these changes.

This PR fixes that so maybe have a deeper look on the code changes.

rosa:
Hey @mhenrixon, I'm back at catching up with PRs. Let me look at this one again with fresh eyes. **2** was already done in a different PR: https://github.com/rails/solid_queue/blob/546864f1d618fb3c5a855d690dc19d2202c5d408/app/models/solid_queue/semaphore.rb#L43

Although the problem there was the opposite of what you're describing: https://github.com/rails/solid_queue/issues/456. A job that was being blocked without having to be blocked. 

I need to look at the remaining 5. I think 1 is still expected behaviour, but I need to look more carefully at the other 4.

mhenrixon:
For me it doesn't really matter anymore. I developed pgbus instead: https://pgbus.zoolutions.llc/

rosa:
Oh, wow, looks very nice! 🤩

I'll finish up this one regardless. Thank you!

rosa:
Merged https://github.com/rails/solid_queue/pull/761, which includes some of the changes here plus other fixes, and released v1.5.0 with it. Closing this one. Thanks again!

PR #690, opened 2025-11-30
Update README.md with MariaDB `SKIP LOCKED` version information

chapmajs:
This PR notes that `SKIP LOCKED` is supported on MariaDB 10.6 and newer, and that one may need to set `config.solid_queue.use_skip_locked = false` for MariaDB versions < 10.6. I ran into this when setting up `solid_queue` on Slackware64 15.0, which is shipping MariaDB 10.5 series still.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ end
 
 ### High performance requirements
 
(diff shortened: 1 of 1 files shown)


rosa:
Oh, great catch! Thanks a lot @chapmajs!

PR #691, opened 2025-12-02
Fix spacing for `database.yml` in `README.md`

danielvdao:
I noticed the spacing was off when configuring multiple databases for the `development` environment.


rosa:
Ohhh, thank you! Good catch!

PR #700, opened 2026-01-10
Fix NoMethodError when status.exitstatus is nil

rickychilcott:
When a worker process is terminated by certain signals (e.g., SIGTERM, SIGKILL), status.exitstatus returns nil, causing a NoMethodError due to operator precedence in the conditional.

This fix:
1. Adds parentheses to ensure proper evaluation order
2. Uses .to_i to safely convert nil to 0, preventing the NoMethodError

Fixes the error: NoMethodError: undefined method '>' for nil

### DIFF
--- a/lib/solid_queue/fork_supervisor.rb
+++ b/lib/solid_queue/fork_supervisor.rb
@@ -38,7 +38,7 @@ module SolidQueue
         pid, status = ::Process.waitpid2(-1, ::Process::WNOHANG)
         break unless pid
 
-        if (terminated_fork = process_instances.delete(pid)) && !status.exited? || status.exitstatus > 0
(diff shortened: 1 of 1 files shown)


rickychilcott:
Thanks for merging!

Since deploying Solid Queue 1.3.0 I noticed this crop up when I did a deploy on Render. It looks like Render sends a SIGTERM and then sends a SIGKILL after that, if the process isn't running (you can read more at https://render.com/docs/deploys#sequence-of-events).

My guess it that this code was a problem on SIGTERM, but I didn't fully instrument to see which signal was causing the issue.

I was surprised to learn that [Process#exitstatus](https://docs.ruby-lang.org/en/2.2.0/Process/Status.html#method-i-exitstatus) could return nil

PR #708, opened 2026-01-27
Support :sql schema format in the install generator

### LINKED ISSUE #621: bin/rails solid_queue:install generate queue_schema.rb instead of structure.sql when config.active_record.schema_format equal :sql
In my app, I use the SQL schema format. When I generate configs using solid_queue:install, it generates queue_schema.rb, which cannot be read because the app does not use the default schema format.
I fixed this issue as follows:
1. Run rails solid_queue:install
2. Comment out config.active_record.schema_format = :sql in my app configs
3. Run rails db:prepare
4. Uncomment config.active_record.schema_format = :sql in my configs
5. Run rails db:migrate:queue

After that, I got the queue_structure.sql that I needed. It feels a little bit tricky, so maybe it would be better to generate queue_structure.sql when ```rails  solid_queue:install``` is run?

ringe:
This patch will make it easier to use `solid_queue` when the app is set up with `:sql` schema format because of unsupported SQL features.

Fixes #621

* Install generator uses schema or structure template based on `active_record.schema_format`
* Added `queue_structure.sql` template based on the existing `queue_schema.rb` template

### DIFF
--- a/lib/generators/solid_queue/install/install_generator.rb
+++ b/lib/generators/solid_queue/install/install_generator.rb
@@ -6,9 +6,15 @@ class SolidQueue::InstallGenerator < Rails::Generators::Base
   def copy_files
     template "config/queue.yml"
     template "config/recurring.yml"
-    template "db/queue_schema.rb"
+    template "db/queue_#{db_format}"
     template "bin/jobs"
     chmod "bin/jobs", 0755 & ~File.umask, verbose: false
+  end
+
+  private
+
+  def db_format
+    Rails.application.config.active_record.schema_format == :sql ? "structure.sql" : "schema.rb"
   end
 
   def configure_adapter_and_database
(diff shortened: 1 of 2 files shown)


ringe:
I've built a script to run any of the "Tests (3.1, sqlite, rails_7_2)" approaches.

The failing tests are not failing because of my patch here, the tests are just not very stable.
I re-run the tests in question and they pass. Re-run the whole file, complete pass.
Sometimes I get skips.

Can I do anything to make the tests more stable?

rosa:
Thank you!

> The failing tests are not failing because of my patch here, the tests are just not very stable.
> I re-run the tests in question and they pass. Re-run the whole file, complete pass.
Sometimes I get skips.
> Can I do anything to make the tests more stable?

Hey! Yes, out of the 3, only one test was actually failing, and that's a flaky test already on my radar. The other two failures were Docker container starting errors from GitHub actions, it seems. 

I'll try to review this this week, but I'm not sure if I'll get to it :crossed_fingers:

rosa:
Sorry for the delay @ringe, I couldn't get to this until this week. 

Unfortunately, this is not going to work as it is, because your structure.sql file is not compatible with PostgreSQL. The advantage of `queue_schema.rb` is that it is RDBMS-agnostic; Rails makes sure to generate the proper SQL statements to use in your DB. A `structure.sql` file is going to change based on multiple external factors. 

I think to support this, the structure.sql file can't be provided as a template; it should be generated as part of the install based on the app's configuration.

rosa (changes requested):
Moving to "changes requested" as this approach won't work because of the reason in https://github.com/rails/solid_queue/pull/708#issuecomment-3922205642

PR #721, opened 2026-03-03
Change default polling_interval to 1 sec

hachi8833:
I think the current `polling_interval: 0.1` for queue.yml is too aggressive so I'm sending the following PR to make the default less aggressive:

```diff
# lib/generators/solid_queue/install/templates/config/queue.yml
     - queues: "*"
       threads: 3
       processes: <%%= ENV.fetch("JOB_CONCURRENCY", 1) %>
-      polling_interval: 0.1
+      polling_interval: 1
```

Actually I encountered several race-conditions with the `polling_interval: 0.1`  when querying the DB (SQLite3) from my dashboard in fly.io.

Hope this change would be helpful for future users 🙏

### DIFF
--- a/lib/generators/solid_queue/install/templates/config/queue.yml
+++ b/lib/generators/solid_queue/install/templates/config/queue.yml
@@ -6,7 +6,7 @@ default: &default
     - queues: "*"
       threads: 3
       processes: <%%= ENV.fetch("JOB_CONCURRENCY", 1) %>
-      polling_interval: 0.1
+      polling_interval: 1
 
 development:
   <<: *default


ghiculescu:
> when querying the DB (SQLite3) from my dashboard in fly.io.

I'm seeing similar issues on Fly.io, on postgres. I've set up https://github.com/ghiculescu/fly-io-replication and have been chatting to a member of their team to try and debug if it's a Fly issue. Changing the polling interval to 1 has made the issues go away for me too.

If you're open to it, I'd love if you could email me (alex@tanda.co) so I can connect you to the Fly engineer I'm chatting to - they would probably really benefit from more examples. I'll also send them a link to this PR.

rosa:
Hey @hachi8833, thanks a lot for this. You're right, it's a bit aggressive. I'm going to merge this one but possibly change the default as well to 1, for folks not using the configuration file 🤔 I'll do that when I release a major version, though.

rosa:
And sorry for the delay too, I'm catching up with PRs now.

airblade:
Does changing the polling default to 1 mean that all jobs now have to wait up to 1s before starting?  I'm wondering about the effect on model broadcasts to the UI.

rosa:
It depends on how often they arrive. The worker pauses for 1 second if there aren't any new jobs before polling again, but if there are new jobs, it continues polling. You can, of course, change the default to the previous polling interval easily in the configuration.

airblade:
Thank you, that's useful to know!

(I'm trying to reduce sqlite database-is-locked errors which I get in solid queue from time to time when I deploy.)

PR #727, opened 2026-03-31
Remove trapping QUIT on Windows

andyundso:
Windows does not support trapping `QUIT`, but everything else. To keep the constant I suggest to implement an `if` switch.


rosa (approved):
Hey @andyundso, thanks for this, and sorry for the delay! I'm catching up with PRs now. I've made a small tweak and added a note to the `README`, and this is ready to go. Thank you!

andyundso:
Hi @rosa, no worries about the delay, thanks for looking at the PR :)

PR #730, opened 2026-04-05
Improve DISTINCT queries on PostgreSQL via recursive CTE

bubiche:
Attempt to improve performance issues with PostgreSQL as stated [here](https://github.com/rails/solid_queue/blob/176721e33f542e07923fe02964cd55d2c18b4389/README.md?plain=1#L329-L349). Follow up of https://github.com/rails/solid_queue/pull/397

PostgreSQL lacks MySQL's Loose Index Scan, so `DISTINCT` queries on `queue_name` degrade to full index scans on large tables. The recursive CTE walks the B-tree index jumping between distinct values.

I have encapsulate the database specific logic in `Record::LooseDistinct` so the rest can just use `distinct_values_of`, which uses recursive CTE to emulate loose index scan on PostgreSQL and normal `DISTINCT` query for MySQL/SQLite.

This should help with:
- Worker polling with prefixes/wildcards (in `QueueSelector`). We have had situations where `solid_queue_ready_executions` exploded even though it's normally small due to issues in other parts of the system and this would have helped.
- Mission Control (in `Queue.all`), this should solve rails/mission_control-jobs#160

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,7 @@ FROM solid_queue_ready_executions
 WHERE queue_name LIKE 'beta%';
 ```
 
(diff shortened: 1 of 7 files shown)


rosa:
Hey @bubiche, thanks for this, and sorry for the delay, I have been working on completely unrelated stuff for a few months and I'm finally catching up with PRs. 

I'm ok with introducing this, but the code needs to look different. A `distinct_values_of` method that takes a `like_conditions` argument (as a first-class parameter) is oddly specific and stands out very much with respect to other RDBMS-specific code here. Let me look into rewriting this in a better way.

Issue #198, opened 2024-04-04
silence_polling not working


anandbait:
I have added `config.solid_queue.silence_polling = true` in my `application.rb`
I have restarted my server after that.
But still I am getting logs as following
```
solid_queue_1  |   SolidQueue::Pause Pluck (1.4ms)  SELECT "solid_queue_pauses"."queue_name" FROM "solid_queue_pauses"
solid_queue_1  |   TRANSACTION (0.6ms)  BEGIN
solid_queue_1  |   SolidQueue::ReadyExecution Pluck (2.1ms)  SELECT "solid_queue_ready_executions"."job_id" FROM "solid_queue_ready_executions" ORDER BY "solid_queue_ready_executions"."priority" ASC, "solid_queue_ready_executions"."job_id" ASC LIMIT $1 FOR UPDATE SKIP LOCKED  [["LIMIT", 3]]
solid_queue_1  |   TRANSACTION (0.6ms)  COMMIT
solid_queue_1  |   SolidQueue::Pause Pluck (1.4ms)  SELECT "solid_queue_pauses"."queue_name" FROM "solid_queue_pauses"
solid_queue_1  |   TRANSACTION (0.5ms)  BEGIN
solid_queue_1  |   SolidQueue::ReadyExecution Pluck (2.0ms)  SELECT "solid_queue_ready_executions"."job_id" FROM "solid_queue_ready_executions" ORDER BY "solid_queue_ready_executions"."priority" ASC, "solid_queue_ready_executions"."job_id" ASC LIMIT $1 FOR UPDATE SKIP LOCKED  [["LIMIT", 3]]
solid_queue_1  |   TRANSACTION (0.5ms)  COMMIT
solid_queue_1  |   SolidQueue::Pause Pluck (1.1ms)  SELECT "solid_queue_pauses"."queue_name" FROM "solid_queue_pauses"
solid_queue_1  |   TRANSACTION (0.5ms)  BEGIN
solid_queue_1  |   SolidQueue::ReadyExecution Pluck (1.8ms)  SELECT "solid_queue_ready_executions"."job_id" FROM "solid_queue_ready_executions" ORDER BY "solid_queue_ready_executions"."priority" ASC, "solid_queue_ready_executions"."job_id" ASC LIMIT $1 FOR UPDATE SKIP LOCKED  [["LIMIT", 3]]
solid_queue_1  |   TRANSACTION (0.5ms)  COMMIT
[...]

PR #732, opened 2026-04-09
Generalize silence_polling to silence_queries

### LINKED ISSUE #210: Generalize `silence_polling` to `silence_queries`

diogovernier:
Closes #210. Supersedes #389.

## Summary

Implements the `silence_queries` approach [approved by @rosa in #198](https://github.com/rails/solid_queue/issues/198#issuecomment-2048122748): a single unified config that silences all of Solid Queue's internal SQL logging, including polling, heartbeats, concurrency maintenance, process pruning, and scheduler dynamic task reloading.

`silence_polling` is preserved as a backward-compatible alias, as [suggested by @namolnad in #210](https://github.com/rails/solid_queue/issues/210).

## Motivation

`silence_polling` only covers worker and dispatcher polling queries. In practice, heartbeat `UPDATE`s (every 60s per process), semaphore expiration, blocked execution unblocking, and dead process pruning are equally noisy, especially in development or when running with `RAILS_LOG_LEVEL=debug`.

Multiple users have reported this ([#210](https://github.com/rails/solid_queue/issues/210), [#389 comments](https://github.com/rails/solid_queue/pull/389), workarounds via custom log-level overrides and monkeypatches), and @rosa indicated in #389 that she'd prefer a single unified toggle over accumulating per-feature silencing flags:

> "I was hoping we could get #210 done instead of having a bunch of different options to silence different parts."

## Approach

- Extracts a shared `with_silenced_queries` helper in `AppExecutor` (which is already included by every class that touches the DB)
- `Poller#with_polling_volume` now delegates to the shared helper instead of inlining the silencing logic
- All periodic internal operations are wrapped: heartbeats, concurrency maintenance, process pruning, scheduler dynamic task reloading, and metadata reload
- One-time boot queries (e.g., `schedule_tasks`, `fail_orphaned_executions`) are deliberately **not** silenced. They can be useful for debugging deployment issues and only run once

## What changed

| File | Change |
|------|--------|
| `lib/solid_queue.rb` | `mattr_accessor :silence_queries` replaces `:silence_polling`; backward-compat alias methods added |
| `lib/solid_queue/app_executor.rb` | New `with_silenced_queries` helper (shared by all call sites) |
| `lib/solid_queue/processes/poller.rb` | `with_polling_volume` now delegates to `with_silenced_queries` |
| `lib/solid_queue/processes/registrable.rb` | Heartbeat timer and `reload_metadata` wrapped |
| `lib/solid_queue/dispatcher/concurrency_maintenance.rb` | `expire_semaphores` and `unblock_blocked_executions` wrapped |
| `lib/solid_queue/supervisor/maintenance.rb` | `prune_dead_processes` wrapped |
| `lib/solid_queue/scheduler/recurring_schedule.rb` | `reschedule_dynamic_tasks` wrapped |
| `README.md` | Docs updated to reference `silence_queries`; mentions `silence_polling` alias |
| `test/unit/worker_test.rb` | Existing polling tests updated + new heartbeat silencing tests + backward-compat test |
| `test/unit/dispatcher_test.rb` | Existing polling tests updated to use new config name |

## Backward compatibility

Existing `config.solid_queue.silence_polling = true/false` continues to work. The engine initializer calls `SolidQueue.public_send("silence_polling=", value)`, which hits our custom setter and delegates to `silence_queries=`. Getter and predicate methods also delegate. No breaking changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,7 +385,7 @@ All the options available to Active Record for multiple databases can be used he
 
 ### Other configuration settings
 
-_Note_: The settings in this section should be set in your `config/application.rb` or your environment config like this: `config.solid_queue.silence_polling = true`
+_Note_: The settings in this section should be set in your `config/application.rb` or your environment config like this: `config.solid_queue.silence_queries = true`
 
 There are several settings that control how Solid Queue works that you can set as well:
 - `logger`: the logger you want Solid Queue to use. Defaults to the app logger.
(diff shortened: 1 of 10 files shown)


rosa:
Hey @diogovernier, thanks a lot for this! 

In the end, I got this PR into Rails: https://github.com/rails/rails/pull/56899, so you just need to set `sql_notifications: false` in the DB configuration for Solid Queue. 

I think this is simpler than trying to do it in the gem so I'm going to close this one in favour of that Rails setting. 

Thanks again!

PR #733, opened 2026-04-09
Fix flaky Puma plugin tests caused by port collision

diogovernier:
## Summary

Related to #602.

Puma plugin tests bind to a hardcoded port (9222). When consecutive tests run, the OS may still hold the previous port in TCP TIME_WAIT state, causing `EADDRINUSE` that cascades into false test failures.

This replaces the fixed port with dynamic allocation via `TCPServer`, so each test gets a guaranteed available port.

To be precise, this PR should fix this flaky test, for instance: https://github.com/rails/solid_queue/actions/runs/24190892956/job/70607570480?pr=732

## What changed

| File | Change |
|------|--------|
| `test/integration/puma/plugin_testing.rb` | Replace hardcoded port 9222 with `TCPServer.new("127.0.0.1", 0)` dynamic allocation |

### DIFF
--- a/test/integration/puma/plugin_testing.rb
+++ b/test/integration/puma/plugin_testing.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "test_helper"
+require "socket"
 
 module PluginTesting
   extend ActiveSupport::Concern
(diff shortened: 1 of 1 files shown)


rosa (approved):
Thank you!

PR #739, opened 2026-04-20
Detect updates to dynamic recurring tasks on poll

### LINKED ISSUE #738: Updates to dynamic recurring tasks aren't picked up until the scheduler restarts
When dynamic tasks enabled, only two cases are handled on each poll:
- Creating new recurring task
- Deleting existing recurring task

The third case, when task updated is not handled. If dynamic task's schedule changed, arguments, queue_name, priority, class_name, etc., the running scheduler keeps firing against the pre-update version until the process is restarted.

**Example:**

```ruby
task = SolidQueue.schedule_recurring_task(
  "platform-import-schedule-14",
  class: "PlatformlmportJob",
  args: [{ company_id: 14 }],
  schedule: "0 4 * * * Europe/Kyiv" # 04:00 EEST
)

# scheduler picks it up on the next poll

task.update!(schedule: "0 8 * * * Europe/Kyiv") # 08:00 EEST

# expected: jobs start at 08:00 EEST
# actual:   scheduler keeps the original "04:00 EEST" cadence until restart
```

Image

cupatea:
Fixes #738 

Scheduler::RecurringSchedule#reschedule_dynamic_tasks previously only detected created and deleted dynamic tasks. Updates to an existing task's were ignored, so the scheduler kept running against the pre-update version until restart.

**Approach:**

Each entry in scheduled_tasks is now a ScheduledRecurringTask struct that pairs the `Concurrent::ScheduledTask` handle with the source `RecurringTask` AR record:

```ruby
ScheduledRecurringTask = Struct.new(:task, :handle) do
  delegate :key, :updated_at, to: :task
  delegate :cancel, to: :handle
end
```

Add new step`reschedule_dynamic_tasks` that replaces any tasks with `updated_at` differs from the cached ones.

### DIFF
--- a/lib/solid_queue/scheduler/recurring_schedule.rb
+++ b/lib/solid_queue/scheduler/recurring_schedule.rb
@@ -4,6 +4,11 @@ module SolidQueue
   class Scheduler::RecurringSchedule
     include AppExecutor
 
+    ScheduledRecurringTask = Struct.new(:task, :handle) do
+      delegate :key, :updated_at, to: :task
+      delegate :cancel, to: :handle
+    end
+
     attr_reader :scheduled_tasks
 
     def initialize(static_tasks, dynamic_tasks_enabled: false)
@@ -34,7 +39,11 @@ module SolidQueue
     end
 
     def schedule_task(task, run_at: task.next_time)
-      scheduled_tasks[task.key] = schedule(task, run_at: run_at)
+      scheduled_tasks[task.key] = ScheduledRecurringTask.new(task, schedule(task, run_at: run_at))
+    end
+
+    def unschedule_task(key)
+      scheduled_tasks.delete(key)&.cancel
     end
 
     def unschedule_tasks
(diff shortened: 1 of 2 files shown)


cupatea:
Hey @rosa, during release testing with Solid Queue dynamic tasks (#553), QA identified an issue that seems to have been missed in the initial implementation.
We’ve applied the same fix as a patch on our side.

rosa:
Hey @cupatea, thank you so much and sorry for the delay! I'm finally catching up with PRs. 

This one was on purpose, though. [The public, documented API](https://github.com/rails/solid_queue#scheduling-and-unscheduling-recurring-tasks-dynamically) to manage dynamic recurring tasks is: 
```
SolidQueue.schedule_recurring_task(..)
SolidQueue.unschedule_recurring_task(..)
```

That is, to update a dynamic task, you need to unschedule and schedule it again. This is to keep the code simpler. 
Because of this, I'm closing this PR, but thank you so much in any case!

PR #740, opened 2026-04-20
Add Rake task to validate complete recurring configuration

J-Verz:
This PR is my attempt at a solution for the problems described in #722. It adds a Rake task `validate_recurring_config` which loads the recurring task configuration (not only for the current RAILS_ENV) and asserts whether every task definition is valid. If any task is invalid, it exits with a non-zero code and prints the validation errors to stdout. If all tasks are valid, it prints 'All recurring tasks are valid' and exits with code 0.

One problem it still has is that it only works for configurations split by Rails environment; if you define tasks in the top-level (like `recurring_with_invalid.yml` in the dummy app), you get an exception. I'm open for suggestions how to fix that.

There are also no tests yet, let me know if you would like me to add those.

### DIFF
--- a/lib/solid_queue/configuration.rb
+++ b/lib/solid_queue/configuration.rb
@@ -70,7 +70,7 @@ module SolidQueue
     end
 
     def valid_recurring_config?
-      invalid_tasks = Hash.new {|hash, key| hash[key] = [] }
+      invalid_tasks = Hash.new { |hash, key| hash[key] = [] }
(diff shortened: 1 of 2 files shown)


rosa:
Hey @J-Verz! Thanks a lot for this! In the end, I ended up merging #745, which validates the whole config and not just recurring tasks. I'm closing this one then, but thanks a lot for contributing! 🙏🏻

PR #741, opened 2026-04-29
Ignore closed-terminal log errors in Puma plugin shutdown

### LINKED ISSUE #737: Puma plugin: Solid Queue processes persist after terminal close due to Errno::EIO blocking shutdown
## Summary

When a terminal window is closed while running a Rails server with the Solid Queue Puma plugin enabled, the Solid Queue supervisor and worker processes are not cleaned up and continue holding the port, preventing the server from being restarted.

## Steps to Reproduce

```bash
rails new repro --api
cd repro
RAILS_ENV=production SOLID_QUEUE_IN_PUMA=1 ./bin/rails s
# Close terminal window (do not use Ctrl-C)
lsof -i :3000
RAILS_ENV=production SOLID_QUEUE_IN_PUMA=1 ./bin/rails s # Fails because port is in use
```

## Expected Behavior

All Solid Queue processes exit when the terminal is closed and the port is freed.

## Actual Behavior

The Solid Queue supervisor and worker processes remain running and hold the port. The server cannot be restarted without manually killing them:

```
$ lsof -i :3000
COMMAND   PID USER   FD   TYPE  DEVICE SIZE/OFF NODE NAME
ruby    72516 brad    9u  IPv4  ...      0t0  TCP localhost:hbci (LISTEN)
ruby    72521 brad    9u  IPv4  ...      0t0  TCP localhost:hbci (LISTEN)
ruby    72522 brad    9u  IPv4  ...      0t0  TCP localhost:hbci (LISTEN)
ruby    72523 brad    9u  IPv4  ...      0t0  TCP localhost:hbci (LISTEN)
```

## Root Cause

When the terminal is closed, the PTY is torn down. The `monitor_puma` thread in the Puma plugin correctly detects that Puma has died by observing `Process.ppid` change to 1, and calls `log` before sending `Process.kill(:INT, $$)`. However, `log_writer.log` attempts to write to STDOUT, which raises `Errno::EIO` because the PTY no longer exists. This exception bubbles up and prevents `Process.kill` from ever being reached, leaving the supervisor and its children running indefinitely.

Because the terminal is already closed, this error is silently swallowed. Adding explicit error logging to a file revealed the cause:

> [2026-04-14 11:44:12 -0400] Failed to write to Puma log: Errno::EIO Input/output error @ rb_sys_fail_on_write - <STDOUT> "Detected Puma has gone away, stopping Solid Queue.."

The relevant code in `lib/puma/plugin/solid_queue.rb`:

```ruby
def monitor(process_dead, message)
  loop do
    if send(process_dead)
      log message          # <-- raises Errno::EIO, Process.kill never reached
      Process.kill(:INT, $$)
      break
    end
    sleep 2
  end
end

def log(...)
  log_writer.log(...)    # <-- Errno::EIO when PTY is torn down
end
```

## Fix

Rescue `Errno::EIO` in the `log` method so that the kill signal is always reached:

```ruby
def log(*args)
  log_writer.log(*args)
rescue Errno::EIO
  # Terminal is gone, ignore logging errors during shutdown
end
```

Note: There may be other relevant errors that should be handled, such as `Errno::EPIPE` or `Errno::EBADF`, but I haven't managed to produce them in my testing.

## Environment

- macOS, Ubuntu (with tmux)
- Puma 8.0.0
- solid_queue 1.4.0

afurm:
## Summary

Fixes #737.

When the Puma plugin monitor detects that Puma has gone away, it logs a shutdown message before sending `INT` to stop Solid Queue. If the controlling terminal has already been closed, Puma's log writer can raise `Errno::EIO`, preventing the shutdown signal from being sent.

This makes plugin logging tolerate `Errno::EIO` so shutdown can continue.

## Changes

- Rescue `Errno::EIO` from the Puma plugin `log` helper
- Add a unit test proving the monitor still sends the shutdown signal when logging fails

## Validation

- `TARGET_DB=sqlite RBENV_VERSION=3.4.4 bundle _2.5.16_ exec ruby -Itest test/unit/puma_plugin_test.rb`
- `TARGET_DB=sqlite RBENV_VERSION=3.4.4 bundle exec ruby -Itest test/integration/puma/plugin_async_test.rb`
- `TARGET_DB=sqlite RBENV_VERSION=3.4.4 bundle exec ruby -Itest test/integration/puma/plugin_fork_test.rb`
- `RBENV_VERSION=3.4.4 bundle exec rubocop lib/puma/plugin/solid_queue.rb test/unit/puma_plugin_test.rb`

### DIFF
--- a/lib/puma/plugin/solid_queue.rb
+++ b/lib/puma/plugin/solid_queue.rb
@@ -129,5 +129,8 @@ Puma::Plugin.create do
 
     def log(...)
       log_writer.log(...)
+    rescue Errno::EIO
+      # The controlling terminal can disappear before the monitor thread shuts
+      # down the child process. Keep shutdown moving even when logging cannot.
     end
 end
--- a/test/unit/puma_plugin_test.rb
+++ b/test/unit/puma_plugin_test.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "puma/plugin/solid_queue"
+
+class PumaPluginTest < ActiveSupport::TestCase
+  class ClosedTerminalLogWriter
+    def log(...)
+      raise Errno::EIO
+    end
+  end
+
+  test "monitor still stops the process when shutdown logging fails" do
+    plugin = Puma::Plugins.find("solid_queue").new
+    plugin.instance_variable_set(:@log_writer, ClosedTerminalLogWriter.new)
+
+    plugin.stubs(:puma_dead?).returns(true)
+    Process.expects(:kill).with(:INT, Process.pid)
+
+    plugin.send(:monitor, :puma_dead?, "Detected Puma has gone away, stopping Solid Queue...")
+  end
+end


afurm:
I saw the `Tests (3.4, sqlite, rails_8_1)` job fail once in `RecurringTasksTest#test_enqueue_and_process_periodic_tasks` with one recurring `StoreResultJob` ending as `"completed"` instead of the configured `"custom_status"`.

I tried to reproduce the exact matrix and seed locally:

```sh
BUNDLE_GEMFILE=$PWD/gemfiles/rails_8_1.gemfile TARGET_DB=sqlite RBENV_VERSION=3.4.4 bundle exec ruby -Itest test/integration/recurring_tasks_test.rb --seed 33880
BUNDLE_GEMFILE=$PWD/gemfiles/rails_8_1.gemfile TARGET_DB=sqlite RBENV_VERSION=3.4.4 bundle exec bin/rails test --seed 33880
```

Both passed locally. The full suite result was 230 runs, 1304 assertions, 0 failures, 2 skips.

This looks unrelated to this PR’s Puma plugin logging change, so I’m treating it as a likely flaky/order-sensitive recurring scheduler failure and will rerun CI.

Issue #405, opened 2024-11-08
Increased memory use when using `bin/jobs` to start Solid Queue?


jp524:
I'm working on a project that was using Solid Queue v0.7.0 and using `bundle exec rake solid_queue:start` to start the process. When I updated to v1.0.0 last week, I also changed the command to start Solid Queue to use `bin/jobs` per current documentation.

I noticed that the memory use of the associated Heroku app was greatly increased. I wasn't sure if this was due to updating the gem or changing the start command, so I tried changing back the command from `bin/jobs` to the Rake task and have noticed a decrease in memory use, per screenshot below.

![Screenshot 2024-11-08 at 09 11 55](https://github.com/user-attachments/assets/8a180ce6-8561-4364-8041-997f38b29c5d)

I read in the discussion on #343 that using `bin/jobs` is supposed to help with memory use thanks to eager loading, so I'm a bit puzzled by this situation.

rosa:
Huh, that's indeed puzzling! May I ask what's your Solid Queue configuration? And whether you have [`config.rake_eager_load`](https://guides.rubyonrails.org/configuring.html#config-rake-eager-load) set to `true` by any chance?

jp524:
Here's my Solid Queue configuration:
```yml
# config/queue.yml
default: &default
  dispatchers:
    - polling_interval: 1
      batch_size: 500
  workers:
    - queues: [ login, background, default ]
      threads: 3
      processes: 1
      polling_interval: 0.1

development:
 <<: *default

test:
 <<: *default

staging:
 <<: *default

production:
 <<: *default

# config/recurring.yml
one:
  class: JobOne
  schedule: every 5 minutes
two:
  class: JobTwo
  schedule: every 75 minutes
solid_queue_cleanup:
  command: "SolidQueue::Job.clear_finished_in_batches"
  schedule: every day at 3 am
  queue: background
```

I don't have `config.rake_eager_load` set at all so it defaults to false I believe.

rosa:
Thanks so much for sending that. 

> I don't have `config.rake_eager_load` set at all so it defaults to false I believe.

Yes, that's right. I'm quite puzzled in any case, about this behaviour... But I wonder if, in your case, eager loading is causing that extra memory to be used in the beginning 😕  Would you be up to do a quick test if you can do that without running into trouble with your Heroku instance? That'd be setting 
```
config.rake_eager_load = true
```
[...]

Issue #330, opened 2024-09-07
Need advice on Solid Queue's memory usage


yjchieng:
Ruby: 3.3.4
Rails: 7.2.1
Solid Queue: 0.7.0, 0.8.2

--------------

I run a Rails App on AWS EC2 instance with 1G of memory.
I notice the solid queue process takes up 15-20% of the instance's memory, which becomes the single largest process by memory usage.

--------------

What I checked:

1) Check memory usage by start/stop supervisorctl 
(I use it to manage my solid queue process)

stop supervisorctl - free memory 276MB
start supervisorctl - free memory 117MB
> It increases 159MB

2) Stop supervisorctl service, and run "solid_queue:start" directly
Trying to see if this is something related to supervisor.

before solid_queue:start - free memory 252MB
after solid_queue:start - free memory 109MB
> It increases 143MB

3) Then I notice there is a latest version.
I upgraded to 0.8.2 (was 0.7.0).

stop supervisorctl - free memory 220MB
start supervisorctl - free memory 38MB
> It increases 182MB

--------------

I need some advise:

1) Is 150-200MB the minimum requirement to run "solid_queue:start"?
2) Is there any setting/feature that I can switch off to reduce memory usage?
3) Is there any setting that I can limit the maximum memory usage?

And, thanks a lot for making this wonderful gem. :)

rosa:
Hey @yjchieng, thanks for opening this issue! 🙏 I think it depends a lot on your app. A brand new Rails app seems to use around 74.6MB memory for me after booting (without Solid Queue, just running Puma). I think the consumption you're seeing is from all the processes together and not just the supervisor, measuring free memory before starting the supervisor and after, as the supervisor will fork more processes. Are you running multiple workers or just one? I think reducing the number of workers there would help. Another thing that might help is using `bin/jobs`, which preloads the whole app before forking, but the gains there are usually quite modest.

rosa:
There might also be something else going on because the only changes from version 0.7.0 to 0.8.2 were for the installing part of Solid Queue; nothing was changed besides the initial installation, so the memory footprint shouldn't have changed. I imagine there is other stuff going on in your AWS instance at the same time that might consuming memory as well.

Focus-me34:
Up 🆙🔥 
[...]

PR #744, opened 2026-05-06
Memoize Runnable#mode to stop allocating a String + StringInquirer pe…

s01ipsist:
# Memoize `Runnable#mode` to stop allocating per poll

## Summary

`SolidQueue::Processes::Runnable#mode` re-runs `(@mode || DEFAULT_MODE).to_s.inquiry`
on every call, allocating a fresh `String` and `ActiveSupport::StringInquirer`.
Because the poll loops invoke `running_as_fork?` / `running_async?` /
`running_inline?` (all of which call `#mode`) from `shutting_down?` on every
iteration, every dispatcher / worker / scheduler tick produces redundant
allocations for a value that never changes after the supervisor sets it.

This PR restores the pre-960694e8 semantic of "compute the inquiry once,
cache it" by moving the inquiry into a custom `mode=` writer and switching
the reader to memoize the default fallback.

## How this regressed

The current per-call allocation was introduced in
[`960694e8`](https://github.com/rails/solid_queue/commit/960694e8dbf42fb87ab2f6ad27aabd54f62d4052)
("Extract Supervised concern from Runnable"). Before that commit, the mode
was set once in `#start`:

```ruby
def start(mode: :supervised)
  @mode = mode.to_s.inquiry
  ...
end

attr_reader :mode
```

After the refactor, the reader was replaced with:

```ruby
attr_writer :mode

def mode
  (@mode || DEFAULT_MODE).to_s.inquiry
end
```

The intent of the change (allowing the writer to be called separately) is
preserved by the patch — the inquiry is just done at write time.

## The diff

```ruby
-    attr_writer :mode
+    def mode=(value)
+      @mode = (value || DEFAULT_MODE).to_s.inquiry
+    end
…
-      def mode
-        (@mode || DEFAULT_MODE).to_s.inquiry
-      end
+      def mode
+        @mode ||= DEFAULT_MODE.to_s.inquiry
+      end
```

Behaviour is identical: `instance.mode = :inline` still results in
`instance.mode.inline?` returning true, and an unset mode still defaults to
`:async`. The only change is *when* the inquiry wrapping happens — once at
write time instead of once per read.

## Reproduction harness

[`solid-queue-idle-memory-repro`](https://github.com/s01ipsist/solid-queue-idle-memory-repro)
is two minimal Rails 8.1 apps (one with `--skip-solid`, one with the
in-Puma SQ plugin), run side-by-side at zero traffic with summed-RSS
sampling and SIGURG-triggered `ObjectSpace.dump_all`. The README walks
through methodology, results, and caveats; the validation patch in
`app-sq-idle/config/initializers/sq_runnable_memoize.rb` applies the
equivalent of this PR as a monkeypatch and was used to confirm the fix
lands cleanly before this PR was opened.

Headline numbers from the harness:

- `app-async` (control): RSS flat at idle.
- `app-sq-idle` (default config): ~30–50 kB/min RSS growth at idle.
- After 60 minutes of idle, `runnable.rb:36` shows up in `heapy diff`
  with **376 retained Strings** and **367 retained StringInquirers** in
  the dispatcher process alone (per hour).
- With this patch applied as a monkeypatch and `GC.start(full_mark: true)`
  before each heap dump, the line disappears from the diff entirely.

The bigger-picture observation (~3 GB/month allocator-fragmentation drift
from poll-loop churn) is best addressed by process recycling, which is
already proposed as #262 — this PR isn't trying to fix that, just to
remove one specifically-Solid-Queue-side contributor.

## Related issues

This is a small, narrow fix. The broader memory-growth reports in #262,
#330, and #405 likely have additional contributors (allocator
fragmentation under the polling architecture, cumulative AR query cache
growth, etc.) that aren't addressed here. But this is one allocation
hotspot that's clearly attributable to a Solid Queue source line and has
a trivial fix.

## Methodology note

For anyone else doing heap diffs on poll-loop processes: `heapy diff`
will dramatically over-report retention unless you call
`GC.start(full_mark: true, immediate_sweep: true)` immediately before
`ObjectSpace.dump_all`. Without that, young-generation objects allocated
since the last GC appear in both snapshots and look like leaks. With it,
real retention shows up. (My initial diffs against this codebase pointed
at `ActiveSupport::Notifications::Fanout` and `ActiveRecord::Relation`
retention; both turned out to be young-gen artifacts that GC cleanly.)

## Tests

Full test suite passes against SQLite (`227 runs, 1299 assertions, 0
failures, 0 errors`). No new tests added — the existing dispatcher /
worker / scheduler tests already exercise the modes that read this
value.

### DIFF
--- a/lib/solid_queue/processes/runnable.rb
+++ b/lib/solid_queue/processes/runnable.rb
@@ -4,7 +4,9 @@ module SolidQueue::Processes
   module Runnable
     include Supervised
 
-    attr_writer :mode
+    def mode=(value)
+      @mode = (value || DEFAULT_MODE).to_s.inquiry
+    end
 
     def start
       run_in_mode do
@@ -33,7 +35,7 @@ module SolidQueue::Processes
       DEFAULT_MODE = :async
 
       def mode
-        (@mode || DEFAULT_MODE).to_s.inquiry
+        @mode ||= DEFAULT_MODE.to_s.inquiry
       end
 
       def run_in_mode(&block)


rosa (approved):
Nice, thank you @s01ipsist! And so sorry for the delay, I'm finally catching up on PRs this week.

PR #745, opened 2026-05-15
add bin/jobs check to validate config before starting

alexanderadam:
Hello,

this is yet another approach for the config issue (#722).
If you have a typo in `recurring.yml` (i.e. `'0 * * * ^'` instead of `'0 * * * *'`) the supervisor boots, reads the bad config, prints the error to a stdout nobody's listening on, and then exits.
`systemctl start` succeeds.
There are no workers running and no error in the journal.
You don't find out until something that should have been processed wasn't.

This PR adds `bin/jobs check` (and `solid_queue:check` for people who'd rather `rake` it 'til they make it) that runs the existing `Configuration` validations for the current Rails env and exits non-zero if anything is bananas.

The one wrinkle: the thread-pool validator pokes at `SolidQueue::Record.connection_pool`, which raises on hosts that don't have DB credentials (CI etc). Added a `skip_db_checks` option so the `check` path can swallow that one specific error and validate everything else.
So supervisor startup keeps the strict behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Solid Queue can be used with SQL databases such as MySQL, PostgreSQL, or SQLite,
   - [Threads, processes, and signals](#threads-processes-and-signals)
   - [Database configuration](#database-configuration)
   - [Other configuration settings](#other-configuration-settings)
+  - [Validating the configuration](#validating-the-configuration)
 - [Lifecycle hooks](#lifecycle-hooks)
 - [Errors when enqueuing](#errors-when-enqueuing)
 - [Concurrency controls](#concurrency-controls)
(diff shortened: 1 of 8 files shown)


rosa:
Thank you so much, @alexanderadam! I tweaked this a bit after merging #753, still need to do a couple of changes before merging. 

Sorry for the delay! I'm finally catching up with PRs.

alexanderadam:
No worries. Feel free to tell me any suggestions. 
Also @lxxxvi told me that he'd like to have it even stricter(?) or so.

I can't remember 😅 but maybe he has some written feedback on that.

rosa:
Oh! Maybe it was related to this https://github.com/rails/solid_queue/pull/661?

rosa:
The tests failing are flaky. I need to fix those. I'm going to merge this and then I might do some changes on top, for example to consolidate all warnings together so they work in the same way 🤔 but this is perfect, thanks so much, Alex!

PR #747, opened 2026-05-30
Accept optional job argument in stopping?

thoughtless:
Rails commit [6d1c401dc7 ("Pass job instance to stopping?")](https://github.com/rails/rails/commit/6d1c401dc7c73e2a39ffbd42a22678cbbcf23d17) changed `ActiveJob::Continuable#checkpoint!` to call `queue_adapter.stopping?(self)`, passing the running job so adapters can decide whether to checkpoint based on it.

This caused errors such as:

    ArgumentError: wrong number of arguments (given 1, expected 0)

The new argument is ignored. Solid Queue still relies solely on the worker shutdown flag. But the signature is now compatible with both the new and old Rails calling conventions.


thoughtless:
As far as I can tell, the failing CI tests are flaky and not caused by this change.

rosa:
Sorry for the delay here! I'm finally finding time to catch up with the pile of open PRs.

rosa (approved):

PR #749, opened 2026-06-13
docs: use single splat in `key` argument for `limits_concurrency`

G-Rath:
Using a double splat means that an error will be raised if more than two non-keyword arguments are passed to the job, whereas a single splat will eat additional arguments of either type.

I think this makes for a better doc since people copy-and-pasting are hopefully less likely to end up with a timebomb that could explode in production


rosa (approved):
Thank you @G-Rath!

PR #750, opened 2026-06-17
Automatically use Rails or a default timezone for the schedule

andyjeffries:
Rather than have SolidQueue act on the server configured time, this change allows you to use either the Rails application's current timezone (by default if in a Rails app) or a specified default.


andyjeffries:
Maybe I'm wrong, but the build failures don't feel anything to do with this change?

rosa:
Hey @andyjeffries, thanks a lot for this, and sorry for the delay! I'm finally catching up with PRs. This makes sense, I'm going to tweak the code a bit and will be merging it shortly.

PR #753, opened 2026-06-24
Warn instead of raising when the connection pool is smaller than the thread pool

### LINKED ISSUE #736: There should be a way to bypass the connection pool vs worker threads check
Currently, Solid Queue doesn't boot in case the ActiveRecord connection pool is smaller than "<number-of-worker-threads> + 2" because of the following check: https://github.com/rails/solid_queue/blob/176721e33f542e07923fe02964cd55d2c18b4389/lib/solid_queue/configuration.rb#L91-L96

Though I can see an argument for making this an advisory limitation, there doesn't appear to be any reason for a hard cap.

We are currently monkey patching Solid Queue to remove this limitation via the following in our `bin/jobs` file:
```ruby
module SolidQueue
  class Configuration
    def ensure_correctly_sized_thread_pool
    end
  end
end
```

This allows us to run our "parallel_io" queue with 150 worker threads but only 15 database connections without issues.

This queue mostly handles OpenAI API calls which typically take 6+ seconds to complete. So while 150 worker threads sounds like a lot, it's all just a bunch of slow HTTP requests and the overall throughput is quite low.

Allowing a lower connection pool number than worker threads would also resolve: https://github.com/rails/solid_queue/issues/627

There is also the async scheduler work done in https://github.com/rails/solid_queue/pull/728 which may provide an even more efficient solution to high IO workloads, but adding a configuration statement or argument to `bin/jobs` to allow arbitrary connection pool vs worker thread sizes should be a super simple improvement.

ajaynomics:
## What

Solid Queue refuses to boot when the database connection pool is smaller than the worker thread pool, raising a `ConfigurationError`. This PR makes the check advisory: Solid Queue boots and logs a warning instead.

## Why

The hard cap rules out configurations that work well in practice. The reporter in #736 runs an I/O-bound queue — 150 threads making slow HTTP calls against a deliberately small pool — where each thread holds a connection only briefly. Today, booting that setup requires monkey-patching the check away.

As suggested in #736, this replaces the hard failure with a warning, which the reporter confirmed fits the need. (The issue also floated a config flag to bypass the check; an unconditional warning is simpler and adds no new configuration surface.)

## How

- Drop `ensure_correctly_sized_thread_pool` from the validation chain, so `Configuration#valid?` reflects only whether the configuration can boot.
- Add `Configuration#warn_about_undersized_thread_pool`, which logs the advisory through `SolidQueue.logger`.
- Call it from `Supervisor.start`, just after the `valid?` check. The `bin/jobs` CLI and both Puma plugin modes — forked and async — start through `Supervisor.start`, so the warning fires once on every boot path.

The warning text matches the previous error, so existing log filters and operator expectations still hold.

## Tests

- `async_supervisor_test`: booting with more threads than connections logs the warning; booting with enough connections stays silent. (Removing the call from `Supervisor.start` fails the first test.)
- `configuration_test`: an undersized pool no longer makes the configuration invalid.

Closes #736.

### DIFF
--- a/lib/solid_queue/configuration.rb
+++ b/lib/solid_queue/configuration.rb
@@ -6,7 +6,6 @@ module SolidQueue
 
     validate :ensure_configured_processes
     validate :ensure_valid_recurring_tasks
-    validate :ensure_correctly_sized_thread_pool
 
     class Process < Struct.new(:kind, :attributes)
       def instantiate
@@ -69,6 +68,13 @@ module SolidQueue
       mode.fork? || @options[:standalone]
     end
 
+    def warn_about_undersized_thread_pool
+      if (db_pool_size = SolidQueue::Record.connection_pool&.size) && db_pool_size < estimated_number_of_threads
+        SolidQueue.logger.warn "Solid Queue is configured to use #{estimated_number_of_threads} threads but the " +
+          "database connection pool is #{db_pool_size}. Increase it in `config/database.yml`"
+      end
+    end
+
     private
       attr_reader :options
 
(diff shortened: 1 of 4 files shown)


rosa (approved):
Thanks @ajaynomics!

PR #757, opened 2026-07-17
Add --only-recurring to run just the recurring scheduler

### LINKED ISSUE #484: Better scheduler configuration
Hello 👋,

Currently, there’s no configuration option to specify if a scheduler should be started or not, this is decided if a recurring task file is given or not.

In our case, we would like to have only one scheduler for better isolation. To do that, we have to use the following configuration:
```
production:
  workers:
    - processes: 0
      threads: 0
```

This will explicitly tell solid_queue to not boot anything apart from the supervisor and since we also give it a recurring task file, it will also start a scheduler.

This workaround works fine for us but I feel like there could be something more intuitive. If you think the same, I can try to come up with a solution. 👍

rafael-pissardo:
## Summary
- Adds `--only-recurring` / `SOLID_QUEUE_ONLY_RECURRING` so `bin/jobs` can run only the recurring scheduler (no workers or dispatchers).
- Mirrors the existing `--skip-recurring` opt-out and replaces the empty workers/dispatchers config workaround.
- Documents the option in the README.

Fixes #484

## Test plan
- [x] `bin/rails test test/unit/configuration_test.rb test/unit/cli_test.rb`
- [ ] Manually: `bin/jobs --only-recurring` registers only a Scheduler process
- [ ] Manually: `SOLID_QUEUE_ONLY_RECURRING=true bin/jobs` behaves the same

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,8 @@ bin/jobs -c config/calendar.yml
 
 You can also skip the scheduler process by setting the environment variable `SOLID_QUEUE_SKIP_RECURRING=true`. This is useful for environments like staging, review apps, or development where you don't want any recurring jobs to run. This is equivalent to using the `--skip-recurring` option with `bin/jobs`.
 
+To run **only** the scheduler (no workers or dispatchers)—for example to isolate recurring tasks on a dedicated process—set `SOLID_QUEUE_ONLY_RECURRING=true` or use the `--only-recurring` option with `bin/jobs`.
+
 This is what this configuration looks like:
 
 ```yml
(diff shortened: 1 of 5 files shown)


rosa:
Thanks @rafael-pissardo! I think this could work, but I'd name the option `only-recurring`, to match `skip-recurring`. Could you also squash all commits into one?

rafael-pissardo:
Updated per feedback:
- Renamed to `--only-recurring` / `SOLID_QUEUE_ONLY_RECURRING` to match `--skip-recurring`
- Squashed into a single commit on top of current `main`

rosa:
@rafael-pissardo thanks a lot! 🙏🏻

Issue #751, opened 2026-06-24
Supervisor remains alive with missing worker process; due scheduled jobs stop being dispatched until restart


sapandiwakar:
## Description

We hit a production incident where Solid Queue stopped processing due scheduled jobs even though Mission Control still showed active workers with recent heartbeats.

Restarting the Solid Queue jobs process immediately fixed the issue.

This appears to be a supervisor/dispatcher/process-liveness issue rather than an application job failure: the process table was missing one expected worker before restart, the remaining workers continued heartbeating, scheduled jobs accumulated, and the supervisor did not appear to replace the missing worker or prune it as dead.

## Environment

- solid_queue: 1.4.0
- Rails: 8.1.3
- Ruby: 3.4.9
- Queue DB: SQLite
- App DB: PostgreSQL
- Mode: fork
- Started via `bin/jobs`
- Deployment: Kamal container
- Worker config:

```yaml
dispatchers:
  - polling_interval: 1
    batch_size: 500

workers:
  - queues: [high_priority, medium_priority, default]
    threads: 3
    processes: 1
    polling_interval: 0.1

  - queues: [high_priority_transport, medium_priority_transport, default_transport]
    threads: 6
    processes: 3
    polling_interval: 0.1
```

Effective production env:

```text
JOB_CONCURRENCY=1
JOB_THREADS=3
TRANSPORT_JOB_CONCURRENCY=3
TRANSPORT_JOB_THREADS=6
DB_POOL_SIZE=40
```

## Expected behavior

The supervisor should keep the configured process set alive:

- 1 Dispatcher
- 1 Scheduler
- 1 Supervisor(fork)
- 4 Workers

If a worker exits, the supervisor should replace it or mark/prune it clearly.

Due scheduled jobs should be moved from `solid_queue_scheduled_executions` to `solid_queue_ready_executions` and then picked up by workers.

## Actual behavior

Mission Control showed only 3 workers before restart:

```text
worker 416 PID 229965
worker 417 PID 236467
worker 418 PID 236473
```

All three had recent heartbeats.

At the same time:

- Scheduled jobs accumulated: 264 scheduled jobs, many delayed by around 20+ hours
- In progress jobs: 0
- Blocked jobs: 0
- Some transport jobs had recently finished, but old delayed scheduled jobs remained stuck
- Restarting the jobs container restored the expected process shape:

```ruby
{
  "Dispatcher" => 1,
  "Scheduler" => 1,
  "Supervisor(fork)" => 1,
[...]

PR #760, opened 2026-07-20
Replace forked processes that fail to boot

sapandiwakar:
## What

Track each forked process until its boot callbacks finish. If a child does not report readiness within a configurable startup timeout, terminate it and let the existing reap-and-replacement path replace that configured process.

This applies only to fork mode. The default startup timeout is five minutes.

## Why

`ForkSupervisor` currently replaces children only when `waitpid2(..., WNOHANG)` reports that they exited. A child that remains alive while blocked during boot or registration is never returned by `waitpid2`, so the supervisor can retain an unusable child indefinitely without replacing it.

For a dispatcher, this means the process can remain present while never reaching its polling loop. Due scheduled jobs then remain in `solid_queue_scheduled_executions` instead of being moved to ready executions.

This addresses the concrete recovery gap observed in #751. It does not attempt to detect a process that successfully booted and later stopped making progress.

## Reproduction

A deterministic local reproduction is to block a dispatcher in an existing startup hook:

```ruby
SolidQueue::Dispatcher.on_start { sleep 60 }
```

Start Solid Queue in fork mode with a dispatcher configured.

On `main`, the dispatcher child remains alive but never reaches its polling loop. Because it has not exited, `waitpid2(..., WNOHANG)` does not return its PID and the supervisor never replaces it.

On this branch, the behavior can be observed quickly by temporarily lowering the timeout:

```ruby
SolidQueue.process_startup_timeout = 2.seconds
SolidQueue::Dispatcher.on_start { sleep 60 }
```

The supervisor logs the startup timeout, terminates that dispatcher PID, and starts its replacement. The replacement also times out while the hook remains blocked, demonstrating that each configured slot is continuously supervised.

The regression test uses the same failure shape by blocking a worker during registration. It verifies that the stalled child is terminated and replaced while a healthy sibling keeps the same PID.

## How

- Create a readiness pipe for each fork.
- Have the child signal readiness after all boot callbacks complete.
- Track each unready PID using a monotonic startup timestamp.
- Terminate only the PID that exceeds the timeout, then reuse the existing replacement path.
- Close readiness descriptors during successful startup, replacement, and shutdown.
- Emit a warning containing the affected child PID.

A timed-out child is killed rather than gracefully terminated because a process blocked during boot cannot reach its run loop to complete a graceful stop.

## Tests

- A fork blocked during registration is terminated and replaced.
- A healthy sibling is not restarted when another child times out.
- A child that becomes ready before the deadline survives a complete supervision cycle.
- Startup timeout warnings include the actual child PID.

Relates to #751.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,7 @@ There are several settings that control how Solid Queue works that you can set a
 - `use_skip_locked`: whether to use `FOR UPDATE SKIP LOCKED` when performing locking reads. This will be automatically detected in the future, and for now, you only need to set this to `false` if your database doesn't support it. For MySQL, that'd be versions < 8; for MariaDB, versions < 10.6; and for PostgreSQL, versions < 9.5. If you use SQLite, this has no effect, as writes are sequential.
 - `process_heartbeat_interval`: the heartbeat interval that all processes will follow—defaults to 60 seconds.
 - `process_alive_threshold`: how long to wait until a process is considered dead after its last heartbeat—defaults to 5 minutes.
+- `process_startup_timeout`: how long a forked process can take to finish booting before the supervisor replaces it—defaults to 5 minutes.
 - `shutdown_timeout`: time the supervisor will wait since it sent the `TERM` signal to its supervised processes before sending a `QUIT` version to them requesting immediate termination—defaults to 5 seconds.
 - `silence_polling`: whether to silence Active Record logs emitted when polling for both workers and dispatchers—defaults to `true`.
 - `supervisor_pidfile`: path to a pidfile that the supervisor will create when booting to prevent running more than one supervisor in the same host, or in case you want to use it for a health check. It's `nil` by default.
--- a/README.md
+++ b/README.md
@@ -404,7 +404,7 @@ There are several settings that control how Solid Queue works that you can set a
 - `use_skip_locked`: whether to use `FOR UPDATE SKIP LOCKED` when performing locking reads. This will be automatically detected in the future, and for now, you only need to set this to `false` if your database doesn't support it. For MySQL, that'd be versions < 8; for MariaDB, versions < 10.6; and for PostgreSQL, versions < 9.5. If you use SQLite, this has no effect, as writes are sequential.
 - `process_heartbeat_interval`: the heartbeat interval that all processes will follow—defaults to 60 seconds.
 - `process_alive_threshold`: how long to wait until a process is considered dead after its last heartbeat—defaults to 5 minutes.
-- `process_startup_timeout`: how long a forked process can take to finish booting before the supervisor replaces it—defaults to 5 minutes.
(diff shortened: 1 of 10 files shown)


sapandiwakar:
Hi @rosa 👋 Thanks for rebasing this onto current `main`!

I looked into the three failing jobs and I'm fairly confident they're pre-existing flakes rather than something introduced by this patch:

- Both SQLite failures (Ruby 3.2 and 4, Rails 7.2) start in `SolidQueue::JobTest#test_try_to_discard_claimed_job` and then cascade into `SQLite3::BusyException` errors. The rebase's parent commit fails the same way without this patch: https://github.com/rails/solid_queue/actions/runs/30112624739
- The MySQL failure (Ruby 3.3, Rails 8.0) is the recurring-task instrumentation timing test you mentioned in #602 as the flakiest one in the suite.

The previous head of this branch was green across all 64 checks with the same substantive change. A rerun of the three jobs should clear them — I don't have permissions to retrigger them myself.

On the approach itself: happy to discuss if you'd prefer a different shape for this, e.g. a smaller supervisor-side contract instead of the readiness pipe. Just let me know and I can adjust the scope.

rosa:
Hey @sapandiwakar, yes! These failures are flaky tests, nothing to do with your changes. 

Nice catch and fix on this bug! I'm making some changes to the solution, mostly around how the callback works. It doesn't seem right to pass over a callback named `on_fork_ready` around to the generic `Supervisor` class and down to the processes. 

I tried to keep `AsyncSupervisor` and `ForkSupervisor` as minimal as possible, with the common part living in `Supervisor`, but this change makes `ForkSupervisor` diverge from that. 

I'll push some changes to see what you think.

rosa:
Ok, I've pushed 3 commits with some changes. I've renamed `process_startup_timeout` to `fork_startup_timeout`, because it applies only to forks (not even to the supervisor, because it doesn't have a parent that can monitor it), whereas the other configuration options, `process_heartbeat_interval` and `process_alive_threshold`  apply to all processes, so I thought it was good to distinguish that. It's an option that doesn't do anything in async mode. 

Then, I've refactored the code considerably, leaving the `Supervisor` class as it was before, and hiding the startup checks in the processes themselves and the `ForkSupervisor` only.

rosa:
I might still do some smaller changes. The config option is the only public change so that one needs to stay when merging, but the rest can change.

rosa:
Going to test this in our app in production to make sure there aren't any issues we might have missed.

sapandiwakar:
Thank you @rosa for the refactoring and for merging this!

The final shape is much cleaner — keeping `Supervisor` untouched and having the processes own their boot guards makes a lot more sense than passing the readiness callback down through the generic class, and `fork_startup_timeout` is indeed a clearer name given it only applies to forked processes.

We've been running the fix in production for a few days (pinned to the earlier revision of this branch) and it's been working well — we'll move to 1.5.1 now that it's released. Thanks again for the quick turnaround on this!

PR #768, opened 2026-07-24
Stop workers when job finalization fails

### LINKED ISSUE #748: `ClaimedExecution` can be stuck in `claimed` state forever when `#finished` raises
## Problem

When the DB is in an unstable state (e.g., a transient network issue), `SolidQueue::ClaimedExecution#perform` sometimes fails to delete the `SolidQueue::ClaimedExecution` and mark `SolidQueue::Job` as finished.

Once this happens (we actually hit it in our production environment), the job remains `claimed` and will never be finished because the owning worker process remains alive.

## Reproduction

<details>
<summary> Reproducible script </summary>
<p>

```ruby
# script.rb
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "sqlite3"
  gem "solid_queue", require: false, github: 'rails/solid_queue'
  gem 'debug'
end

# Setup application

require 'active_job/railtie'
require 'active_record/railtie'
require 'solid_queue'

require "minitest/autorun"

File.delete('solid_queue_repro.sqlite3') if File.exist?('solid_queue_repro.sqlite3')
ENV["DATABASE_URL"] = "sqlite3:solid_queue_repro.sqlite3?timeout=5000"

class TestApp < Rails::Application
  config.eager_load = false
  config.logger = Logger.new(File.open('rails.log', 'w'))
  config.logger.level = Logger::INFO
  config.secret_key_base = "secret_key_base"

  config.active_job.queue_adapter = :solid_queue
  config.solid_queue.silence_polling = true
end

Rails.application.initialize!

# Create tables for Solid Queue
ActiveRecord::Schema.define do
  instance_eval(File.read(SolidQueue::Engine.root.join("lib/generators/solid_queue/install/templates/db/queue_schema.rb")))
end

class TestJob < ActiveJob::Base
  queue_as :default

  def perform
    # do nothing
  end
end

# Mimics DB instability
SolidQueue::ClaimedExecution.prepend(
  Module.new do
    # https://github.com/rails/solid_queue/blob/176721e33f542e07923fe02964cd55d2c18b4389/app/models/solid_queue/claimed_execution.rb#L110-L115
    def finished
      raise 'Intentional failure that mimics a transient DB glitch' if rand < 0.5

      super
    end
  end
)

# Test the behavior under DB instability
class TestTransientDbFailure < ActiveSupport::TestCase
  def setup
    @supervisor_pid = fork do
      config = SolidQueue::Configuration.new(queues: "default", workers: 1)
      SolidQueue::Supervisor.start(configuration: config)
    end
  end

  def teardown
    Process.kill(:TERM, @supervisor_pid)
    Process.waitpid(@supervisor_pid)
  end

  def test_transient_db_failure
    100.times do
      TestJob.set(wait: 2.seconds).perform_later
    end
    assert_equal 100, SolidQueue::ScheduledExecution.count

    # Never finishes
    until SolidQueue::Job.finished.count == 100
      sleep 1
      puts "finished_count: #{SolidQueue::Job.finished.count}, claimed_count: #{SolidQueue::ClaimedExecution.count}"
    end
  end
end
```

</p>
</details>

I would expect Solid Queue to eventually finish every job, but in practice some jobs are stuck in `claimed` forever:

```bash
$ ruby script.rb
---snip---
finished_count: 0, claimed_count: 0
finished_count: 0, claimed_count: 3
finished_count: 46, claimed_count: 54
finished_count: 46, claimed_count: 54
finished_count: 46, claimed_count: 54
finished_count: 46, claimed_count: 54
finished_count: 46, claimed_count: 54
...
```

## Root Cause

When a worker completes the job, it calls [ClaimedExecution#finished](https://github.com/rails/solid_queue/blob/176721e33f542e07923fe02964cd55d2c18b4389/app/models/solid_queue/claimed_execution.rb#L110) to delete the `ClaimedExecution` and mark `Job` finished.
If that transaction raises, the exception ends up in the `on_rejection!` block inside [Pool](https://github.com/rails/solid_queue/blob/176721e33f542e07923fe02964cd55d2c18b4389/lib/solid_queue/pool.rb#L28-L30), and it swallows the exception.

As a result:

* The `ClaimedExecution` is **not** deleted
* The owning worker process is still alive and registered, so the execution is **not** considered orphaned and is never released by any recovery mechanisms
* The job therefore remains `claimed` indefinitely, even though it is completed

## Possible Solutions

I'm listing two possible approaches below, but I'm not sure either of them is ideal.
I'd love to hear your thoughts and discuss this.

### 1. Stop the worker when `ClaimedExecution#finished` fails

One possible solution to this issue is to stop the worker when it fails to complete `ClaimedExecution`.

The following is the PoC for this approach.
https://github.com/rails/solid_queue/compare/main...genya0407:solid_queue:feature/experimental-change-for-fixing-claimed-execution?expand=1

With this change, the previous reproducible script successfully completed all the jobs.

```bash
# Change the gem source
$ cat main.rb | grep 'gem "solid_queue"'
gem "solid_queue", require: false, github: 'genya0407/solid_queue', branch: 'feature/experimental-change-for-fixing-claimed-execution'

$ ruby script.rb
---snip---
finished_count: 0, claimed_count: 0
finished_count: 0, claimed_count: 0
finished_count: 1, claimed_count: 3
finished_count: 6, claimed_count: 3
finished_count: 8, claimed_count: 3
---snip---
finished_count: 99, claimed_count: 0
finished_count: 99, claimed_count: 0
finished_count: 100, claimed_count: 0
```

### 2. Periodic worker restarts

Once the worker is restarted, the stuck claimed executions will be released immediately and eventually retried.
In fact, when we encountered this issue, we restarted the workers by restarting the supervisors, and the jobs were released and re-executed.

So maybe supporting an automatic restart feature based on uptime or the number of processed jobs could mitigate this problem and others.
I believe, aside from this problem, restarting worker processes constantly is a common practice.

rafael-pissardo:
## Summary
- wrap finalization failures while a claimed execution is still present in a dedicated error
- stop the owning worker after that unrecoverable error so shutdown releases the claim for retry
- keep workers running after ordinary job failures and avoid joining from a pool thread

Fixes #748.

## Test plan
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/unit/worker_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/models/solid_queue/claimed_execution_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/integration/jobs_lifecycle_test.rb`
- [x] RuboCop on changed Ruby files

### DIFF
--- a/app/models/solid_queue/claimed_execution.rb
+++ b/app/models/solid_queue/claimed_execution.rb
@@ -1,153 +1,158 @@
 # frozen_string_literal: true
 
-class SolidQueue::ClaimedExecution < SolidQueue::Execution
-  belongs_to :process
+module SolidQueue
+  class ClaimedExecution < Execution
+    belongs_to :process
 
-  scope :orphaned, -> { where.missing(:process) }
+    scope :orphaned, -> { where.missing(:process) }
 
-  class Result < Struct.new(:success, :error)
-    def success?
-      success
-    end
-  end
-
-  # Raised when a job has already run (or failed) but we couldn't update its
-  # claim/finished state because of a transient error. The claim is still held
-  # by a living worker, so it won't be recovered as orphaned unless the worker
-  # is stopped and replaced.
-  class FinalizationError < RuntimeError
-    def initialize(claimed_execution, cause:)
-      super("Failed to finalize claimed execution #{claimed_execution.id} (job #{claimed_execution.job_id}): #{cause.class}: #{cause.message}")
-      set_backtrace(cause.backtrace) if cause.backtrace
-    end
-  end
-
-  class << self
-    def claiming(job_ids, process_id, &block)
-      job_data = Array(job_ids).collect { |job_id| { job_id: job_id, process_id: process_id } }
-
-      SolidQueue.instrument(:claim, process_id: process_id, job_ids: job_ids) do |payload|
-        insert_all!(job_data)
-        where(job_id: job_ids, process_id: process_id).load.tap do |claimed|
-          block.call(claimed)
-
-          payload[:size] = claimed.size
-          payload[:claimed_job_ids] = claimed.map(&:job_id)
-        end
+    class Result < Struct.new(:success, :error)
+      def success?
+        success
       end
     end
 
-    def release_all
-      SolidQueue.instrument(:release_many_claimed) do |payload|
-        includes(:job).tap do |executions|
-          executions.each(&:release)
-
-          payload[:size] = executions.size
-        end
+    # Raised when a job has already run (or failed) but we couldn't update its
+    # claim/finished state because of a transient error. The claim is still held
+    # by a living worker, so it won't be recovered as orphaned unless the worker
+    # is stopped and replaced.
+    class FinalizationError < Processes::UnrecoverableError
+      def initialize(claimed_execution, cause:)
+        super("Failed to finalize claimed execution #{claimed_execution.id} (job #{claimed_execution.job_id}): #{cause.class}: #{cause.message}")
+        set_backtrace(cause.backtrace) if cause.backtrace
       end
     end
 
-    def fail_all_with(error)
-      includes(:job).tap do |executions|
-        return if executions.empty?
+    class << self
+      def claiming(job_ids, process_id, &block)
+        job_data = Array(job_ids).collect { |job_id| { job_id: job_id, process_id: process_id } }
 
-        SolidQueue.instrument(:fail_many_claimed) do |payload|
-          executions.each do |execution|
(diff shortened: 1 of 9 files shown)


rosa:
Thanks a lot @rafael-pissardo! I made some changes mostly around style, this was a nice catch!

genya0407:
@rafael-pissardo 
Thanks for the fix! ❤️

@rosa 
Are there any plans to release this fix?
We still encounter this issue intermittently, so we'd love to see a new release soon!

PR #769, opened 2026-07-27
Use with_connection in DistinctValues

trevorturk:
The new `DistinctValues` module from #730 calls `connection` directly, which causes errors when using `config.active_record.permanent_connection_checkout = :disallowed`. 

This PR wraps the calls in `connection_pool.with_connection` following the pattern from #652.


rosa (approved):
Oh, great catch! Thanks a lot @trevorturk 🙏🏻

PR #770, opened 2026-07-27
Honour TERM received during supervisor boot before forking workers

### LINKED ISSUE #755: Forked worker misses forwarded TERM, keeps claiming jobs until SIGKILL
### Environment

- solid_queue 1.4.0, Rails 7.x, Ruby 3.x
- Workers run as Kubernetes Deployments, scaled by an HPA that reacts to queue depth with an aggressive scale-up (0s stabilization window) so pods can go from "just provisioned" to "asked to terminate" within seconds — either because the HPA scales back down almost immediately after a burst, or because the node backing the pod (we run these workers on Spot-backed node groups) gets reclaimed. A rollout can also mass-replace pods faster than they finish booting. `terminationGracePeriodSeconds` is generous (100s+) specifically to give in-flight jobs a chance to finish, but that budget is wasted if the worker never reacts to the signal at all.

### The race

Solid Queue's supervisor only *acts* on a received signal inside its `supervise` loop, via `process_signal_queue` — that loop runs **after** `start_processes` has already forked workers (`lib/solid_queue/supervisor.rb`):

```ruby
def start
  boot              # registers the supervisor's own trap (enqueue-only)
  run_start_hooks   # can take a while (app boot, cache warmup, etc.)
  start_processes   # forks workers
  launch_maintenance_task
  supervise         # <- first place a queued signal is actually processed
end
```

The supervisor's trap itself only enqueues (`lib/solid_queue/supervisor/signals.rb`):

```ruby
trap(signal) { signal_queue << signal; interrupt }
```

If TERM arrives during `run_start_hooks` (a slow app boot is common), it sits queued until `supervise` runs `perform_graceful_termination`, which forwards TERM to the just-forked workers via `term_forks` (`lib/solid_queue/fork_supervisor.rb`). Because this forwarding happens within microseconds of the fork, it lands **before** the child has run its own `register_signal_handlers` (`lib/solid_queue/processes/supervised.rb`, called from `boot`, i.e. after fork). The freshly-forked worker still holds the **inherited** copy of the supervisor's enqueue-only trap at the moment the signal arrives — so the forwarded TERM gets pushed onto a `signal_queue` that only supervisors ever drain. A worker process never calls `process_signal_queue`, so the signal is silently dropped: `@stopped` is never set, `shutting_down?` stays `false`, and the poller (`lib/solid_queue/processes/poller.rb`) keeps calling `poll` → `ReadyExecution.claim` indefinitely.

The worker keeps claiming and running jobs in a pod that Kubernetes believes is already draining, until the grace period expires and the pod is SIGKILLed. The in-flight claimed job is orphaned; five minutes later another supervisor's maintenance task detects the stale heartbeat and fails it with `SolidQueue::Processes::ProcessPrunedError` — for a job that may have completed or been mid-flight, not one that genuinely crashed.

We confirmed this with production logs across multiple pods, e.g. (paraphrased, timestamps redacted to relative offsets):

```
T+0.000  Started Supervisor(fork) pid: 7        # supervisor's trap already armed
T+0.010  preStop sends TERM to supervisor pid
T+2.900  Started Worker pid: 152                # forked ~3s later, after TERM already queued
T+2.970  Starting perform job                   # worker claims a job AFTER the TERM
T+101.7  Starting perform job                   # STILL claiming new jobs seconds before kill
T+102.0  pod killed, exit code 137 (SIGKILL)     # no graceful shutdown log lines at all
```

Notably absent from the whole window: `Supervisor terminated gracefully`, `Supervisor terminated immediately`, or `Shutdown Worker` — the shutdown path never runs at all for that worker.

There is a second, related manifestation of the same underlying gap: if termination is requested even earlier — while the **app itself** (not just the supervisor) is still booting, before `on_start` hooks have run — nothing in solid_queue's boot path checks for a pending termination, so the boot completes and a worker gets forked into a pod that's already being torn down.

### Related issues

- #589 — "Finished jobs ending up as failed with `ProcessPrunedError`" (open, no root-cause diagnosis)
- #591 — "Need to retry jobs that died with `ProcessPrunedError`" (open, asks for a retry mechanism rather than a fix)
- #422 — closed, similar symptom reported without a fork-timing/signal analysis

None of these identify the fork-timing/signal-forwarding race described above, so we're filing a new issue rather than commenting on an existing one.

### How we worked around it (without patching the gem)

We close the race entirely from the outside, using only public hooks and our own container lifecycle scripts:

1. **Worker announces "I can handle a graceful TERM" via a ready file**, written from an `after_boot` hook (`on_process_start`) — which runs only once the worker's own `TERM → stop` trap is armed. The file's existence is the signal that this specific process (not just "a supervisor is up") is ready to shut down cleanly.
2. **Our preStop hook waits for that ready file before sending TERM** to the supervisor (bounded wait, falls back to a timeout marker if the worker never becomes ready). This guarantees TERM is only forwarded once the worker has already replaced the inherited enqueue-only trap with its own — closing the fork-timing race without touching solid_queue's source.
3. **Liveness probing is kept on the supervisor's pidfile**, not the ready file, so a healthy-but-still-booting supervisor isn't killed by kubelet while waiting for step 1.
4. For the earlier boot-time variant: our preStop hook writes a separate "terminating" marker *immediately*, before waiting for readiness. An `on_start` hook (`on_supervisor_start`, which runs before `start_processes` forks anything) checks that marker and exits before forking a worker at all if termination was already requested.

All of this lives in our app's solid_queue hooks and our container's start/stop scripts — nothing in `lib/solid_queue/` is modified.

### What would help upstream

We think this is worth fixing natively rather than requiring every Kubernetes deployer to reinvent the ready-file dance:

- Have the worker (child) re-arm its signal trap as the very first thing after `fork`, before the parent's `term_forks` has a chance to signal it — or have the supervisor briefly block sending TERM to a just-forked child until it acknowledges it's ready (e.g. via a pipe/socket rather than relying on trap timing).
- Alternatively, expose an official "worker ready" hook/notification (what we built ourselves as a ready file) so operators don't need out-of-band file-based coordination to know when it's safe to forward TERM.
- Document the boot-time gap (TERM received before `on_start` hooks run) and consider having the supervisor check for a pending signal before `start_processes` runs.

Happy to share our hook implementation if useful as a reference.

rafael-pissardo:
## Summary
- Drain the supervisor signal queue before each child is forked, and shut down immediately when a stop signal was already queued during boot/start hooks
- Re-arm `TERM`/`INT`/`QUIT` handlers as the first thing in forked children so a forwarded TERM is not handled by the inherited enqueue-only supervisor trap
- Add a regression test that sends `TERM` while an `on_start` hook is still running and asserts workers never start or claim jobs

Fixes #755.

## Test plan
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/integration/supervisor_boot_signal_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/integration/forked_processes_lifecycle_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/integration/lifecycle_hooks_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/unit/fork_supervisor_test.rb`
- [x] RuboCop on changed files

### DIFF
--- a/lib/solid_queue/processes/runnable.rb
+++ b/lib/solid_queue/processes/runnable.rb
@@ -50,7 +50,13 @@ module SolidQueue::Processes
         case
         when running_as_fork?
           @boot_guard = BootGuards::ForkGuard.new
-          fork(&block).tap { @boot_guard.start }
+          # Re-arm signal handlers as the first thing in the child so a TERM
+          # forwarded by the supervisor right after fork is not handled by the
+          # inherited enqueue-only supervisor trap (see #755).
+          fork do
+            register_signal_handlers
+            block.call
+          end.tap { @boot_guard.start }
         when running_async?
           @boot_guard = BootGuards::NullGuard.new
           @thread = create_thread(&block)
--- a/lib/solid_queue/processes/runnable.rb
+++ b/lib/solid_queue/processes/runnable.rb
@@ -50,13 +50,7 @@ module SolidQueue::Processes
         case
         when running_as_fork?
           @boot_guard = BootGuards::ForkGuard.new
-          # Re-arm signal handlers as the first thing in the child so a TERM
-          # forwarded by the supervisor right after fork is not handled by the
-          # inherited enqueue-only supervisor trap (see #755).
-          fork do
-            register_signal_handlers
-            block.call
-          end.tap { @boot_guard.start }
+          create_fork(&block).tap { @boot_guard.start }
         when running_async?
           @boot_guard = BootGuards::NullGuard.new
           @thread = create_thread(&block)
@@ -70,10 +64,7 @@ module SolidQueue::Processes
       def boot
         SolidQueue.instrument(:start_process, process: self) do
           run_callbacks(:boot) do
-            if running_as_fork?
-              register_signal_handlers
-              set_procline
-            end
+            set_procline if running_as_fork?
           end
         end
 
--- a/lib/solid_queue/supervisor.rb
+++ b/lib/solid_queue/supervisor.rb
@@ -39,11 +39,13 @@ module SolidQueue
       run_start_hooks
 
       start_processes
-      return shutdown if stopped?
 
-      launch_maintenance_task
-
-      supervise
+      if stopped?
+        shutdown
+      else
+        launch_maintenance_task
+        supervise
+      end
     end
 
     def stop
@@ -68,32 +70,30 @@ module SolidQueue
 
       def start_processes
         configuration.configured_processes.each do |configured_process|
-          # Honour signals that arrived during boot / start hooks before forking
-          # any children. Otherwise a queued TERM is only handled in #supervise,
-          # after workers have already been started (see #755).
-          process_signal_queue if standalone?
-          break if stopped?
+          # Honour signals that arrive during boot or start hooks: a queued TERM
+          # stops us here, before starting children, instead of in #supervise,
+          # after all of them have been started
+          break if time_to_stop?
 
           start_process(configured_process)
         end
       end
 
       def supervise
-        loop do
-          break if stopped?
-
-          if standalone?
-            set_procline
-            process_signal_queue
-          end
-
-          unless stopped?
-            check_and_replace_terminated_processes
-            interruptible_sleep(1.second)
-          end
+        until time_to_stop?
+          set_procline
+          check_and_replace_terminated_processes
+          interruptible_sleep(1.second)
         end
       ensure
         shutdown
+      end
+
+      # Process any signals queued while we were busy and report whether
+      # we've been asked to stop
+      def time_to_stop?
+        process_signal_queue
+        stopped?
       end
 
       def start_process(configured_process)
@@ -149,7 +149,10 @@ module SolidQueue
       end
 
       def set_procline
-        procline "supervising #{configured_processes.keys.join(", ")}"
+        # Embedded supervisors don't own their process's title
+        if standalone?
+          procline "supervising #{configured_processes.keys.join(", ")}"
+        end
       end
 
       def sync_std_streams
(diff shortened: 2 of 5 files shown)


lib/solid_queue/processes/runnable.rb
```diff
@@ -50,7 +50,13 @@ def run_in_mode(&block)
         case
         when running_as_fork?
           @boot_guard = BootGuards::ForkGuard.new
-          fork(&block).tap { @boot_guard.start }
+          # Re-arm signal handlers as the first thing in the child so a TERM
+          # forwarded by the supervisor right after fork is not handled by the
+          # inherited enqueue-only supervisor trap (see #755).
+          fork do
+            register_signal_handlers
```
rosa: This feels redundant together with the same call to `register_signal_handlers` right after running the `before_boot` callbacks... (see `#boot`, which runs first thing as part of `block.call` when running as fork). I think one of those needs to go.

lib/solid_queue/supervisor.rb
```diff
@@ -65,7 +67,15 @@ def boot
       end
 
       def start_processes
-        configuration.configured_processes.each { |configured_process| start_process(configured_process) }
+        configuration.configured_processes.each do |configured_process|
+          # Honour signals that arrived during boot / start hooks before forking
```
rosa: This would run for any supervisor even if it doesn't fork (`AsyncSupervisor`), so I don't think it belongs here.

rosa: Ah, no, it affects the `AsyncSupervisor` as well, it's just the comment that's misleading.

rosa (approved):
I pushed a small refactor of your changes, but haven't changed the functionality. Thank you, good catch!

PR #771, opened 2026-07-27
Keep concurrency locks across non-graceful job release

### LINKED ISSUE #735: limits_concurrency bypassed after non-graceful shutdown
## Summary

When a worker is force-killed during shutdown (shutdown timeout exceeded), jobs protected by limits_concurrency can run concurrently after restart.

## Steps to reproduce

```
class SlowJob < ActiveJob::Base
  limits_concurrency key: "slow_job", duration: 5.minutes

  def perform
    sleep 1.hour
  end
end
```

1. Enqueue 3 SlowJob instances
2. Start SolidQueue supervisor in fork mode (worker thread_pool_size: 3)
3. Wait for the first job to start (semaphore acquired, others blocked)
4. Send SIGTERM to the supervisor
5. SolidQueue.shutdown_timeout (default 5s) expires — supervisor force-kills the worker
6. Start a new supervisor
7. Two or more jobs start Performing concurrently, violating the concurrency limit of 1

## Expected behavior

Only one SlowJob runs at a time after restart, same as before the shutdown.

## Actual behavior

Multiple jobs with the same concurrency key run simultaneously after restart.

## Root cause

Supervisor#start calls start_processes (line 39), which starts the dispatcher and workers concurrently. The dispatcher's ConcurrencyMaintenance is initialized with Concurrent::TimerTask.new(run_now: true), so it does run expire_semaphores and unblock_blocked_executions at boot — but in a background thread. Meanwhile, the worker starts polling immediately and can claim multiple jobs before the maintenance thread completes.

## The sequence:

1. Old worker is force-killed mid-job, leaving a stale semaphore in solid_queue_semaphores
2. Release claimed jobs runs, putting the interrupted job back in the ready queue
3. New supervisor starts — dispatcher and workers boot concurrently
4. Dispatcher's maintenance starts in a background thread (Concurrent::TimerTask)
5. Worker starts polling (every 0.1s), claims multiple ready jobs before maintenance has expired the stale semaphore and unblocked blocked executions
6. Concurrency limit is violated

## Observed in production logs

14:38:39 Supervisor wasn't terminated gracefully - shutdown timeout exceeded (5018.5ms)
14:38:39 Release claimed jobs (90.1ms)  size: 1
...
14:51:47 ==> Your service is live
14:51:50 [Job ff2291c7] Performing RefreshDataJob (az4n-8mr2)
14:51:50 [Job b1ddfa0c] Performing RefreshDataJob (6sqe-dvqs)

Both jobs use limits_concurrency key: self (limit 1) but started in the same second after a deploy that triggered a non-graceful shutdown.

## Possible fix

Run ConcurrencyMaintenance#expire_semaphores and #unblock_blocked_executions synchronously during dispatcher boot, before workers start polling. This would ensure stale semaphores from dead processes are cleaned up before any jobs are claimed.

## Environment
- solid_queue 1.4.0
- Rails 8.1
- Ruby 3.4.7
- PostgreSQL 16
- Fork mode supervisor

rafael-pissardo:
## Summary
- Extend a job's concurrency semaphore expiry when a claimed execution is released back to ready, so the slot stays owned across restart
- Skip expiring semaphores whose key still has a ready job, and run concurrency maintenance synchronously at supervisor boot and dispatcher start before workers can poll
- Add regression tests for the ready-job lock race from non-graceful shutdown

Fixes #735.

## Test plan
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/unit/concurrency_maintenance_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/integration/concurrency_controls_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/models/solid_queue/claimed_execution_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/unit/dispatcher_test.rb`
- [x] `TARGET_DB=sqlite bundle exec ruby -Itest test/unit/fork_supervisor_test.rb`
- [x] RuboCop on changed files

### DIFF
--- a/app/models/solid_queue/claimed_execution.rb
+++ b/app/models/solid_queue/claimed_execution.rb
@@ -75,6 +75,10 @@ class SolidQueue::ClaimedExecution < SolidQueue::Execution
   def release
     SolidQueue.instrument(:release_claimed, job_id: job.id, process_id: process_id) do
       unless_already_finalized do
+        # Keep owning the concurrency slot while the job waits to be claimed
+        # again so maintenance cannot expire the lock and unblock another job
+        # for the same key (#735).
+        job.extend_concurrency_lock
         job.dispatch_bypassing_concurrency_limits
         destroy!
       end
--- a/app/models/solid_queue/job/concurrency_controls.rb
+++ b/app/models/solid_queue/job/concurrency_controls.rb
@@ -33,6 +33,12 @@ module SolidQueue
         blocked_execution.present?
       end
 
+      def extend_concurrency_lock
+        return unless concurrency_limited?
+
+        Semaphore.where(key: concurrency_key).update_all(expires_at: concurrency_duration.from_now)
+      end
+
       private
         def concurrency_on_conflict
           job_class.concurrency_on_conflict.to_s.inquiry
--- a/lib/solid_queue/dispatcher/concurrency_maintenance.rb
+++ b/lib/solid_queue/dispatcher/concurrency_maintenance.rb
@@ -6,15 +6,25 @@ module SolidQueue
 
     attr_reader :interval, :batch_size
 
+    class << self
+      def perform(batch_size: SolidQueue::Configuration::DISPATCHER_DEFAULTS[:batch_size])
+        new(nil, batch_size).perform
+      end
+    end
+
     def initialize(interval, batch_size)
       @interval = interval
       @batch_size = batch_size
     end
 
     def start
-      @concurrency_maintenance_task = Concurrent::TimerTask.new(run_now: true, execution_interval: interval) do
-        expire_semaphores
-        unblock_blocked_executions
+      # Run once inline so stale locks are cleared before workers forked alongside
+      # this dispatcher can claim jobs (see #735). The timer then handles the
+      # periodic follow-up without racing that first pass via run_now.
+      perform
+
+      @concurrency_maintenance_task = Concurrent::TimerTask.new(run_now: false, execution_interval: interval) do
+        perform
       end
 
       @concurrency_maintenance_task.add_observer do |_, _, error|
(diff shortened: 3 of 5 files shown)


rosa:
Hey @rafael-pissardo, thanks for this, but I'm not sure these changes are compatible with the way concurrency controls work. What they fix is intentional behaviour. 

> Extend a job's concurrency semaphore expiry when a claimed execution is released back to ready, so the slot stays owned across restart

The expiration is based on the max duration specified in the job's concurrency settings. This means that after that time has passed, we can no longer guarantee that other jobs will run. This is the case no matter what the job is doing. If the job went back to ready, that means that either the worker crashed, or it was gracefully terminated with the job going over the shutdown timeout, or something similar. If the lock expires during that time, we consider that to be the time the job takes to run, and it counts towards the concurrency duration. The concurrency duration is a self-healing mechanism to avoid having jobs blocked forever. The trade-off is that in some cases, depending on the jobs and the duration configured, it might happen that some jobs overlap. This is preferable to having jobs blocked forever. 

> Skip expiring semaphores whose key still has a ready job, and run concurrency maintenance synchronously at supervisor boot and dispatcher start before workers can poll

This is the same idea. If we have ready jobs when their locks have already expired, then they need to overlap with other jobs, that's the way it's intended to work. From the README: 

> Since something can happen that prevents the first job from releasing the semaphore and unblocking the next job (for example, someone pulling a plug in the machine where the worker is running), we have the duration as a failsafe. **Jobs that have been blocked for more than duration are candidates to be released**, but only as many of them as the concurrency rules allow, as each one would need to go through the semaphore dance check. This means that **the duration is not really about the job that's enqueued or being run, it's about the jobs that are blocked waiting**, or about the jobs that would get discarded while the semaphore is closed.

Thank you!

PR #774, opened 2026-07-30
Reschedule dynamic recurring tasks when they are updated

### LINKED ISSUE #738: Updates to dynamic recurring tasks aren't picked up until the scheduler restarts

wintan1418:
Fixes #738

### Problem

When dynamic recurring tasks are enabled, the scheduler's polling cycle (`Scheduler::RecurringSchedule#reschedule_dynamic_tasks`) only handles two cases:

- `schedule_created_dynamic_tasks` — rows whose key isn't scheduled yet
- `unschedule_deleted_dynamic_tasks` — scheduled entries whose key is no longer in the DB

A task whose key exists in both places falls through both checks, so any update to its `schedule`, `class_name`, `arguments`, `queue_name`, etc. is silently ignored: the scheduler keeps firing against the pre-update definition until the process restarts.

```ruby
task = SolidQueue.schedule_recurring_task("import", class: "ImportJob", schedule: "0 4 * * *")
# scheduler picks it up on the next poll

task.update!(schedule: "0 8 * * *")
# expected: fires at 08:00 from now on
# actual:   keeps firing at 04:00 until the scheduler restarts
```

### Fix

Keep the previously loaded snapshot of dynamic tasks across each polling cycle and compare every reloaded task's `updated_at` against it. Any task that changed gets its in-memory scheduled entry cancelled and rescheduled from the fresh record, mirroring what a delete + create with a polling cycle in between would have done.

### Testing

Added a regression test that starts a scheduler with a dynamic task on an hourly cadence, updates it to `every second` while the scheduler is running, and asserts jobs get enqueued without a restart. The test fails on `main` and passes with this change.

### DIFF
--- a/lib/solid_queue/scheduler/recurring_schedule.rb
+++ b/lib/solid_queue/scheduler/recurring_schedule.rb
@@ -48,8 +48,10 @@ module SolidQueue
 
     def reschedule_dynamic_tasks
       wrap_in_app_executor do
+        previous_tasks = dynamic_tasks.index_by(&:key)
         reload_dynamic_tasks
         schedule_created_dynamic_tasks
+        reschedule_updated_dynamic_tasks(previous_tasks)
         unschedule_deleted_dynamic_tasks
       end
     end
@@ -71,6 +73,16 @@ module SolidQueue
 
       def schedule_created_dynamic_tasks
         RecurringTask.dynamic.where.not(key: scheduled_tasks.keys).each do |task|
+          schedule_task(task)
+        end
+      end
+
+      def reschedule_updated_dynamic_tasks(previous_tasks)
+        dynamic_tasks.each do |task|
+          previous_task = previous_tasks[task.key]
+          next if previous_task.nil? || previous_task.updated_at == task.updated_at
+
+          scheduled_tasks[task.key]&.cancel
           schedule_task(task)
         end
       end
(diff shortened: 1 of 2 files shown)


rosa:
Hey @wintan1418, thanks for this, but this is expected behaviour. Check this comment: https://github.com/rails/solid_queue/pull/739#issuecomment-5022493765

wintan1418:
Thanks @rosa — makes sense, and sorry I missed the earlier discussion on #739. Closing this one. Since it's tripped a few people now, I've opened #777 adding a short note to the README's dynamic tasks section spelling out that in-place updates aren't picked up and that unschedule + schedule again is the way to update a task.

PR #777, opened 2026-07-30
Document how to update dynamic recurring tasks

wintan1418:
Follow-up to #774 and #739, related to #738.

Updating a dynamic recurring task in place (e.g. `task.update!(schedule: ...)`) isn't picked up by a running scheduler — by design, the way to update a task is to unschedule it and schedule it again, as @rosa explained in https://github.com/rails/solid_queue/pull/739#issuecomment-5022493765.

Since this wasn't stated in the README's dynamic tasks section and has now come up a few times (#738 plus two PRs attempting the same fix), this adds a short paragraph spelling it out.


rosa (approved):
Great, thank you!

Issue #763, opened 2026-07-23
Scheduler becomes an un-replaced zombie after host sleep/resume; recurring jobs silently stop


julik:
**Version:** solid_queue 1.3.1 · Rails 8 · Puma plugin (`SOLID_QUEUE_IN_PUMA=true`) · single host · SQLite queue DB

## Summary

After the host sleeps and resumes, the **scheduler** ends up **alive at the OS level but deregistered from `solid_queue_processes`**, with its scheduling thread dead. The fork supervisor only replaces children that *exit*, so it never replaces it, and **no recurring jobs are enqueued again until a manual restart**. Dispatcher and worker recover normally, so ad-hoc jobs keep running and hide the outage.

This is the scenario PR #337 targeted ("a computer going to sleep and then coming back to life"), but the self-heal it added doesn't cover the scheduler here. No existing issue ties the sleep/resume **trigger** to the scheduler-**zombie** outcome end-to-end.

In practice this means that a Rails app running on a laptop that sleeps and then awakens stops running its scheduled tasks until restarted.

## Evidence

Last `calendar_sync` recurring execution: **2026-06-30 19:30 UTC**; nothing for days after, while ad-hoc jobs kept running and no `FailedExecution` rows were created.

**`solid_queue_processes` has no Scheduler row** — only supervisor/dispatcher/worker (the latter two re-forked after the sleep). But `ps` shows the scheduler still alive:

```
61362 solid-queue-fork-supervisor(1.3.1): supervising 61365, 43988, 43991
61365 solid-queue-scheduler(1.3.1): scheduling calendar_sync,...
```

Pid 61365 is alive and supervised but has no DB registration → zombie.

Log around the sleep boundary:

```
Deregister Scheduler   process_id: 5403, pid: 61365, last_heartbeat_at: "2026-06-30T19:11:08Z", pruned: true
Error enqueuing recurring task   task: "calendar_sync", at: "2026-06-30T19:12:00Z"
```

After this: **no further `Register Scheduler` and no `Replaced terminated Scheduler`**. Dispatcher/worker get `Replaced terminated …` repeatedly across the same period; the scheduler only ever gets `Shutdown Scheduler`.

## Root cause

1. During sleep no process heartbeats. On resume, the dispatcher's prune sweep deletes the others' stale `Process` rows, including the scheduler's.
2. Sleep freezes rather than kills: the scheduler resumes with its row already pruned, so its next enqueue errors and its scheduling thread dies.
[...]

PR #778, opened 2026-07-30
Stop processes whose heartbeats keep failing past the alive threshold

wintan1418:
Related to #763 (and the same family as #751).

### What I found

The core scenario in #763 — a process whose registration gets pruned while it's still alive — already self-heals on `main`: the next heartbeat raises `RecordNotFound`, `Registrable#heartbeat` clears the process and wakes the run loop, `shutting_down?` returns true via `!registered?`, the fork exits, and the supervisor replaces it. I've added a regression test proving this end-to-end for the scheduler (start a supervisor with a scheduler, delete the scheduler's `solid_queue_processes` row while it's running, assert a fresh registration appears).

### The remaining hole

That self-heal only fires when the heartbeat gets a clean `RecordNotFound`. Heartbeats can also fail without confirming anything about the registration — a dropped DB connection after host sleep/resume, SQLite busy errors (the reporter's setup in #763) — and those errors are only reported via the timer task's observer. The process then keeps running unregistered indefinitely: the supervisor never replaces it because it never exits, and from the outside it's a zombie that stops doing visible work. This matches the reported evidence: an alive, supervised scheduler with no `solid_queue_processes` row and no "Replaced terminated Scheduler" ever logged.

### Fix

Track when heartbeats started failing. Once they've been failing for longer than `SolidQueue.process_alive_threshold` — the exact point at which every other process considers this one dead and prunable — treat it like finding the registration gone: clear the process and wake the run loop so the process stops and the supervisor starts a replacement that can register afresh. Errors are still re-raised so the existing observer reporting is unchanged, and a successful heartbeat resets the tracking, so transient blips don't accumulate.

### Testing

- New unit test: a worker whose heartbeats persistently raise `ActiveRecord::StatementInvalid` shuts down its pool once the failures outlast the alive threshold (fails on `main`, passes with this change).
- New regression test for the existing pruned-registration replacement path (passes on `main` and with this change).
- Full worker/scheduler/supervisor/process-recovery test files green on MySQL.

### DIFF
--- a/app/models/solid_queue/process.rb
+++ b/app/models/solid_queue/process.rb
@@ -21,10 +21,14 @@ class SolidQueue::Process < SolidQueue::Record
 
   def heartbeat
     # Clear any previous changes before locking, for example, in case a previous heartbeat
-    # failed because of a DB issue (with SQLite depending on configuration, a BusyException
-    # is not rare) and we still have the unpersisted value
+    # failed because of a DB issue and we still have the unpersisted value
     restore_attributes
     with_lock { touch(:last_heartbeat_at) }
+  rescue
+    # touch writes the attribute before persisting; don't let a failed
+    # update leave this object claiming a heartbeat that was never persisted
+    restore_attributes
+    raise
   end
 
   def deregister(pruned: false)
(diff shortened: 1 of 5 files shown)


rosa (approved):
Thanks @wintan1418! I pushed some mostly style-related changes. Good catch!

PR #779, opened 2026-07-30
Include the error in fail_many_claimed events and document recovery

### LINKED ISSUE #591: Need to retry jobs that died with ProcessPrunedError
We have some situations where we get ProcessPrunedErrors for jobs that were on killed workers (in Kubernetes). We need a way to have those retried. 

`on_thread_error` seems like it might be appropriate, but it is very unclear what the lifecycle would look like using it since the examples are all just for error captures.

wintan1418:
Closes #591

### Approach

Following the discussion in #591: not automatic retries (poison-pill risk, as @rosa pointed out), but the hook @nhorton asked for, so applications that know their jobs are idempotent can implement their own recovery policy.

It turns out most of the machinery already exists: `ClaimedExecution.fail_all_with` — the path that marks in-flight jobs as failed when their process is pruned or a crashed fork is reaped — already emits `fail_many_claimed.solid_queue` with the affected `job_ids` and `process_ids`. What was missing was the `error` in the payload, so subscribers couldn't distinguish a prune (`ProcessPrunedError`) from other mass failures, and any documentation that this hook exists.

This PR:

- Adds `error` to the `fail_many_claimed` payload (and to its log line, using the existing `formatted_error` convention).
- Extends the existing instrumentation test to assert the error is included.
- Documents the recovery pattern in the README under "Failed jobs and retries": why these jobs aren't retried automatically, and a subscriber example that re-enqueues via `FailedExecution#retry` with an explicit note that the application must bring its own safeguard against retry loops.

No new machinery, no behavior change for anyone not subscribing to the event.

### Testing

Instrumentation and log subscriber test files green on MySQL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Solid Queue can be used with SQL databases such as MySQL, PostgreSQL, or SQLite,
   - [Performance considerations](#performance-considerations)
 - [Failed jobs and retries](#failed-jobs-and-retries)
   - [Error reporting on jobs](#error-reporting-on-jobs)
+  - [Jobs interrupted by non-graceful process death](#jobs-interrupted-by-non-graceful-process-death)
 - [Puma plugin](#puma-plugin)
 - [Jobs and transactional integrity](#jobs-and-transactional-integrity)
 - [Recurring tasks](#recurring-tasks)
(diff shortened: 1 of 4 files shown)


rosa (approved):
Nice, thank you!

PR #781, opened 2026-07-30
Replace terminated forks even if releasing their claimed jobs fails

### LINKED ISSUE #683: SolidQueue stops processing any jobs, but does not actually exit, when DB restarts unless --skip-recurring is set
Steps to reproduce:
- Have some jobs defined, including at least one recurring job.
- Run solidqueue
- Restart the database (for example on a mac, `brew services restart mariadb`, or on linux, `systemctl restart mariadb`)
- Notice that an error appears in the terminal that ran solidqueue, but it doesn't exit (not returned to shell), and no jobs are processed in the future

If you do this again but run with `--skip-recurring`, solidqueue will print the error and then exit. This is useful because when running solidqueue in a process monitor such as systemd, the solidqueue process exiting will cause the monitor to restart it - if the database restarted just once, this means no future jobs are missed.

wintan1418:
Fixes #683

### Problem

When the database restarts under a running fork supervisor, the workers and dispatcher crash and exit (their polling loops raise). The supervisor reaps them and calls `replace_fork`, which runs `release_claimed_jobs_by` — a database call — *before* starting the replacement. With the database still down, that call raises, the exception escapes `check_and_replace_terminated_processes`, and the whole `supervise` loop crashes without ever replacing the terminated forks.

The scheduler, whose per-task enqueue errors are reported without killing the process, survives and reconnects — so after the database comes back, recurring jobs keep being enqueued with no workers left to run them. I reproduced this end-to-end on `main` with `docker compose restart mysql` under `bin/jobs`: a growing backlog of ready executions, worker heartbeats frozen at the restart timestamp, and a supervisor that never recovers. Full chain of evidence — including the recovered crash backtrace and why the process hangs instead of exiting when the `debug` gem is loaded — is written up in a comment on #683.

### Fix

Rescue errors from `release_claimed_jobs_by` in `replace_fork`, report them through `handle_thread_error`, and start the replacement regardless. The database being unreachable at that moment is likely the very reason the fork died, so the replacement can't depend on it.

The terminated fork's claimed jobs aren't lost by skipping the release: its stale process registration is pruned once the database is reachable again, and pruning fails its claimed executions through the normal `ProcessPrunedError` path.

### Testing

- New unit test: `replace_fork` still starts the replacement when releasing claimed jobs raises `ConnectionNotEstablished` (errors without this change, passes with it).
- Live verification of the original scenario: with this change, `bin/jobs` under a `docker compose restart mysql` recovers by itself — replacement workers crash-loop through the boot-timeout path while the database is down, then re-register and drain the backlog as soon as it's back. On `main`, the same scenario ends with a permanently dead supervisor loop and an ever-growing backlog.

### DIFF
--- a/lib/solid_queue/fork_supervisor.rb
+++ b/lib/solid_queue/fork_supervisor.rb
@@ -73,8 +73,16 @@ module SolidQueue
         if terminated_fork = process_instances.delete(pid)
           terminated_fork.mark_as_reaped
           payload[:fork] = terminated_fork
-          error = Processes::ProcessExitError.new(status)
-          release_claimed_jobs_by(terminated_fork, with_error: error)
+
+          begin
+            release_claimed_jobs_by(terminated_fork, with_error: Processes::ProcessExitError.new(status))
+          rescue StandardError => error
+            # The database may be unreachable — likely the same reason the fork
+            # terminated. Starting the replacement can't depend on it: the jobs
+            # claimed by the terminated fork will be failed when its stale
+            # registration is pruned once the database is back.
+            handle_thread_error(error)
+          end
 
           start_process(configured_processes.delete(pid))
         end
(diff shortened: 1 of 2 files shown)


rosa (approved):
Thank you @wintan1418, very nice!

PR #783, opened 2026-08-01
Destroy orphan BlockedExecution rows when the job class no longer resolves

### LINKED ISSUE #784: DelegationError in BlockedExecution release path when a job's class_name no longer resolves
## Summary

`solid_queue` (verified in 1.4.0 and still present in 1.6.0) raises `ActiveSupport::DelegationError: concurrency_limit delegated to job_class, but job_class is nil` from the dispatcher's concurrency-maintenance tick whenever a `SolidQueue::BlockedExecution` row exists for a `SolidQueue::Job` whose `class_name` no longer `safe_constantize`s (e.g. the ActiveJob class was renamed, removed, or namespaced differently between deploys).

There is already a guard for this condition (`Job#concurrency_limited?` returns `false` when `job_class.blank?`), but the **release** path called by the dispatcher does not consult it — it goes straight to `Semaphore.wait(job)`, which then dereferences the missing delegate target.

Result: the dispatcher's `unblock_blocked_executions` timer fires every tick, hits the same orphan rows, and reports the exception on every attempt indefinitely. The blocked rows are never released or cleaned up on their own.

## Environment

- `solid_queue`: reproduced on 1.4.0, still present in 1.6.0 (source verified — the relevant files are byte-identical between 1.4.0 and 1.6.0).
- Ruby: 3.4.2
- Rails / ActiveRecord: 8.1.3.1
- Backend: PostgreSQL (not adapter-specific — the bug is model-layer).

## Reproduction

1. Enqueue an ActiveJob that uses `limits_concurrency to: N, key: ...` and that gets blocked (its concurrency semaphore is at cap), producing at least one `solid_queue_blocked_executions` row.
2. In the next deploy, delete or rename the job class so that `class_name.safe_constantize` returns `nil` for the row created in step 1.
3. Wait for `SolidQueue::Dispatcher::ConcurrencyMaintenance#unblock_blocked_executions` to fire (it runs on a timer).
4. Every tick raises `ActiveSupport::DelegationError` and, because the release transaction rolls back, the same orphan row is picked up again on the next tick — the error will recur forever.

Alternative repro without a real code change: manually update a `solid_queue_jobs.class_name` value to a string that does not resolve, e.g.

```ruby
SolidQueue::Job.find(<blocked job id>).update_columns(class_name: "GoneJob")
```

Then wait for the dispatcher tick.

## Stack trace (from a real production hit)

```
ActiveSupport::DelegationError: concurrency_limit delegated to job_class, but job_class is nil
NoMethodError: undefined method 'concurrency_limit' for nil

solid_queue (1.4.0) app/models/solid_queue/job/concurrency_controls.rb:11:in 'SolidQueue::Job#concurrency_limit'
solid_queue (1.4.0) app/models/solid_queue/semaphore.rb:86:in 'SolidQueue::Semaphore::Proxy#limit'
solid_queue (1.4.0) app/models/solid_queue/semaphore.rb:58:in 'SolidQueue::Semaphore::Proxy#attempt_creation'
solid_queue (1.4.0) app/models/solid_queue/semaphore.rb:46:in 'SolidQueue::Semaphore::Proxy#wait'
solid_queue (1.4.0) app/models/solid_queue/semaphore.rb:10:in 'SolidQueue::Semaphore.wait'
solid_queue (1.4.0) app/models/solid_queue/blocked_execution.rb:65:in 'SolidQueue::BlockedExecution#acquire_concurrency_lock'
solid_queue (1.4.0) app/models/solid_queue/blocked_execution.rb:49:in 'block (2 levels) in SolidQueue::BlockedExecution#release'
...
solid_queue (1.4.0) app/models/solid_queue/blocked_execution.rb:28:in 'SolidQueue::BlockedExecution.release_one'
solid_queue (1.4.0) app/models/solid_queue/blocked_execution.rb:24:in 'SolidQueue::BlockedExecution.release_many'
solid_queue (1.4.0) app/models/solid_queue/blocked_execution.rb:14:in 'SolidQueue::BlockedExecution.unblock'
solid_queue (1.4.0) lib/solid_queue/dispatcher/concurrency_maintenance.rb:40:in 'block in SolidQueue::Dispatcher::ConcurrencyMaintenance#unblock_blocked_executions'
solid_queue (1.4.0) lib/solid_queue/app_executor.rb:7:in 'SolidQueue::AppExecutor#wrap_in_app_executor'
solid_queue (1.4.0) lib/solid_queue/dispatcher/concurrency_maintenance.rb:39:in 'SolidQueue::Dispatcher::ConcurrencyMaintenance#unblock_blocked_executions'
solid_queue (1.4.0) lib/solid_queue/dispatcher/concurrency_maintenance.rb:17:in 'block in SolidQueue::Dispatcher::ConcurrencyMaintenance#start'
```

## Root cause

`SolidQueue::Job::ConcurrencyControls` sets up a delegate:

```ruby
# app/models/solid_queue/job/concurrency_controls.rb
delegate :concurrency_limit, :concurrency_duration, to: :job_class

def job_class
  @job_class ||= class_name.safe_constantize
end

def concurrency_limited?
  concurrency_key.present? && job_class.present?
end
```

`Job#acquire_concurrency_lock` correctly gates on `concurrency_limited?`:

```ruby
def acquire_concurrency_lock
  return true unless concurrency_limited?
  Semaphore.wait(self)
end
```

But `BlockedExecution#acquire_concurrency_lock` (a different method on a different class, invoked from `BlockedExecution#release` in the dispatcher-driven release path) does **not**:

```ruby
# app/models/solid_queue/blocked_execution.rb
def acquire_concurrency_lock
  Semaphore.wait(job)
end
```

`Semaphore.wait` calls `Semaphore::Proxy#limit`, which calls `job.concurrency_limit`, which is the delegate — and blows up when `job_class` is `nil`.

Because this happens inside `BlockedExecution#release`'s transaction, the row is never destroyed and never promoted; the dispatcher rediscovers it on every subsequent tick.

`set_expires_at` in the same file (`self.expires_at = job.concurrency_duration.from_now`, called on `before_create`) has the same latent problem, but only bites at create time — new orphans wouldn't be created here in practice, since the class exists when the block is initially recorded.

## Impact

- Every `SolidQueue::Dispatcher::ConcurrencyMaintenance` tick raises and reports an exception for as long as the orphaned rows exist — days, weeks, indefinitely.
- Ops noise: floods exception dashboards from a source that is entirely internal state, unrelated to any user request.
- Real blocked jobs that share a `concurrency_key` with an orphan may also be starved, because `release_one` locks *one* row per key inside a transaction that then rolls back.
- Any deploy that removes or renames a concurrency-controlled ActiveJob class is a landmine.

## Suggested fix

Two small guards, symmetric with the existing `Job#acquire_concurrency_lock` gate:

1. **`BlockedExecution#release`** — short-circuit and destroy the row when the job class no longer resolves, so the dispatcher stops retrying forever.
2. **`BlockedExecution#set_expires_at`** — fall back to `SolidQueue.default_concurrency_control_period` when the class does not resolve.

julik:
Fixes #784.

## Summary

Fixes a `DelegationError` raised on every `SolidQueue::Dispatcher::ConcurrencyMaintenance` tick when a `BlockedExecution` row references a `Job` whose `class_name` no longer `safe_constantize`s — e.g. because the concurrency-limited ActiveJob class was renamed or removed between deploys.

`Job#acquire_concurrency_lock` already gates on `concurrency_limited?` (which checks `job_class.present?`), but `BlockedExecution#acquire_concurrency_lock` — invoked from the dispatcher-driven release path — went straight to `Semaphore.wait(job)`, which calls `job.concurrency_limit`, which is delegated to a `nil` `job_class`:

```
ActiveSupport::DelegationError: concurrency_limit delegated to job_class, but job_class is nil
  app/models/solid_queue/job/concurrency_controls.rb:11:in 'concurrency_limit'
  app/models/solid_queue/semaphore.rb:86:in 'limit'
  app/models/solid_queue/semaphore.rb:58:in 'attempt_creation'
  app/models/solid_queue/semaphore.rb:46:in 'wait'
  app/models/solid_queue/blocked_execution.rb:65:in 'acquire_concurrency_lock'
  app/models/solid_queue/blocked_execution.rb:49:in 'block (2 levels) in release'
  ...
  lib/solid_queue/dispatcher/concurrency_maintenance.rb:40
```

Because the exception is raised inside `BlockedExecution#release`'s transaction, the row is never destroyed or promoted, and the dispatcher re-picks the same orphans forever — flooding error reporters and starving any legitimately blocked jobs that share the concurrency key.

## Changes

- `BlockedExecution#release`: if `job.job_class` is `nil`, destroy the orphan row and emit `orphaned: true` in the `release_blocked` instrumentation payload. Otherwise fall through to the existing acquire/promote/destroy path.
- `BlockedExecution#set_expires_at`: fall back to `SolidQueue.default_concurrency_control_period` when the class does not resolve (guards the same latent bug at create time).
- `Job#job_class` promoted from private to public so `BlockedExecution` can consult it.
- New unit test covering the orphan release path.

## Test plan

- [x] `bin/rails test test/models/solid_queue/blocked_execution_test.rb` — new test passes.
- [x] `bin/rails test test/models/` — full model suite green (97 runs, 585 assertions, 0 failures).
- [x] CI across mysql/postgres/sqlite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,6 +555,8 @@ It's important to note that after one or more candidate jobs are unblocked (eith
 
 When using `discard` as the behaviour to handle conflicts, you might have jobs discarded for until the `duration` interval if something happens and a running job fails to release the semaphore.
 
+If a job's class no longer exists by the time its concurrency controls are checked—say it was renamed or removed in a deploy while jobs referencing it were still in the queue—the job is marked as failed with a `SolidQueue::Job::ClassMissingError`, so it shows up in [failed jobs](#failed-jobs-and-retries), where it can be retried once the class is back, or discarded. Jobs with a missing class picked up by a worker fail with the same error.
+
 
 For example:
 ```ruby
--- a/app/models/solid_queue/blocked_execution.rb
+++ b/app/models/solid_queue/blocked_execution.rb
@@ -2,6 +2,8 @@
 
 module SolidQueue
   class BlockedExecution < Execution
+    class JobClassMissingError < RuntimeError; end
+
     assumes_attributes_from_job :concurrency_key
     before_create :set_expires_at
 
@@ -48,9 +50,12 @@ module SolidQueue
         transaction do
           if job.job_class.nil?
             # The job's class no longer resolves (renamed/removed between deploys).
-            # Destroy the orphan row so the dispatcher stops retrying it forever.
+            # Mark the job as failed so it surfaces in Mission Control (where it can
+            # be retried once the class is restored, or discarded), and destroy the
+            # orphan row so the dispatcher stops retrying it forever.
+            job.failed_with(JobClassMissingError.new("Job class #{job.class_name.inspect} could not be resolved"))
             destroy!
-            payload[:orphaned] = true
+            payload[:failed] = true
           elsif acquire_concurrency_lock
             promote_to_ready
             destroy!
--- a/app/models/solid_queue/blocked_execution.rb
+++ b/app/models/solid_queue/blocked_execution.rb
@@ -46,7 +46,12 @@ module SolidQueue
     def release
       SolidQueue.instrument(:release_blocked, job_id: job.id, concurrency_key: concurrency_key, released: false) do |payload|
         transaction do
-          if acquire_concurrency_lock
+          if job.job_class.nil?
+            # The job's class no longer resolves (renamed/removed between deploys).
+            # Destroy the orphan row so the dispatcher stops retrying it forever.
+            destroy!
+            payload[:orphaned] = true
+          elsif acquire_concurrency_lock
             promote_to_ready
             destroy!
 
@@ -58,7 +63,8 @@ module SolidQueue
 
     private
       def set_expires_at
-        self.expires_at = job.concurrency_duration.from_now
+        duration = job.job_class ? job.concurrency_duration : SolidQueue.default_concurrency_control_period
+        self.expires_at = duration.from_now
       end
 
       def acquire_concurrency_lock
--- a/app/models/solid_queue/blocked_execution.rb
+++ b/app/models/solid_queue/blocked_execution.rb
@@ -2,8 +2,6 @@
 
 module SolidQueue
   class BlockedExecution < Execution
-    class JobClassMissingError < RuntimeError; end
-
     assumes_attributes_from_job :concurrency_key
     before_create :set_expires_at
 
@@ -49,12 +47,9 @@ module SolidQueue
       SolidQueue.instrument(:release_blocked, job_id: job.id, concurrency_key: concurrency_key, released: false) do |payload|
         transaction do
           if job.job_class.nil?
-            # The job's class no longer resolves (renamed/removed between deploys).
-            # Mark the job as failed so it surfaces in Mission Control (where it can
-            # be retried once the class is restored, or discarded), and destroy the
-            # orphan row so the dispatcher stops retrying it forever.
-            job.failed_with(JobClassMissingError.new("Job class #{job.class_name.inspect} could not be resolved"))
+            job.failed_with(Job::ClassMissingError.for(job))
             destroy!
+
             payload[:failed] = true
           elsif acquire_concurrency_lock
             promote_to_ready
--- a/app/models/solid_queue/claimed_execution.rb
+++ b/app/models/solid_queue/claimed_execution.rb
@@ -115,6 +115,8 @@ module SolidQueue
       end
 
       def execute
+        raise Job::ClassMissingError.for(job) if job.job_class.nil?
+
         ActiveJob::Base.execute(job.arguments.merge("provider_job_id" => job.id))
         Result.new(true, nil)
       rescue Exception => e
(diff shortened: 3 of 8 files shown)


julik:
@rosa any chance getting a look at this?

rosa:
Hey @julik, sorry for the delay! I've been on-call for a couple of weeks, and working on a lot of other stuff, but I'm on vacation this week and catching up with PRs. I have a few in the queue before this one, so it'll take a few days more.

rosa:
Hey @julik! Sorry again for the delay. I'm looking into this one now. I wonder: would it make sense, instead of just discarding these jobs, to mark them as failed instead, so you have the chance of doing something with them? Or discarding if you don't need them, that's always an option.

julik:
I will check - I think anything but having things "occupying a slot" will work

julik:
Done - hope this works ;-)

rosa:
Thanks @julik! I've pushed a couple of small tweaks to your solution, but basically this is it. Thank you!

Issue #422, opened 2024-11-21
`SolidQueue::Processes::ProcessPrunedError` shown in "Failed jobs" and not being released back to their queues


salmonsteak1:
Hey there, the README states this :
> _If processes have no chance of cleaning up before exiting (e.g. if someone pulls a cable somewhere), in-flight jobs might remain claimed by the processes executing them. Processes send heartbeats, and the supervisor checks and prunes processes with expired heartbeats, which will release any claimed jobs back to their queues. You can configure both the frequency of heartbeats and the threshold to consider a process dead. See the section below for this._

From what I understand, this means that when the plug is pulled or the worker is terminated unexpectedly, jobs that are still in the midst of processing by that worker will be released back to their queue the next time the supervisor checks and prunes processes with expired heartbeats. 

However, it seems like I'm getting these `ProcessPrunedError`:
<img width="1700" alt="image" src="https://github.com/user-attachments/assets/bbb7b18e-bc50-4742-a108-f2a23624a39e">

<img width="876" alt="image" src="https://github.com/user-attachments/assets/84ef0da1-0dbe-4c5f-800e-3a14ed12de04">

I believe this happens in the event of the "plug being pulled" (or something similar like forcefully shutting down a kubernetes pod), the supervisor will also be killed and the job will be stuck with the worker. I'm just not sure when/where this `ProcessPrunedError` is being generated, and if there's anything we can do about it for it to be re-enqueued

wflanagan:
I've run into this as well. We implemented a query to find them, and then try to requeue them. But, it seems we've not done this well, and it's turned into a bit of a mess. It seems like when the process is terminated, we should throw them back to the queue to attempt again.

rosa:
@salmonsteak1, oh yes! I forgot to update the README on https://github.com/rails/solid_queue/pull/277. In-flight jobs whose worker terminates abnormally will now be marked as failed, just like if the job failed, so you can inspect them and retry/discard/fix them manually. 

The reason is that the worker might be crashing _because of the job_. For example, a job that has a memory leak causes some monitoring process to kill a worker because of memory consumption. If we simply put the job back into the queue when the process is killed, another worker will pick it up just to be killed and so on. I need to update the README. 

[...]

Issue #313, opened 2024-08-30
unable to retry on SolidQueue::ProcessExitError


InteNs:
hello again!

After updating to V0.6 we naturally started getting failed jobs with `SolidQueue::ProcessExitError` as the error causing the job to fail.

This looked like a perfect candidate of an exception that we should always retry so i tried adding the following to ApplicationJob:
```ruby
retry_on SolidQueue::ProcessExitError, wait: :polynomially_longer, attempts: 5
```

unfortunately it seems that this error class is not known to the application:
```
app/jobs/application_job.rb:16:in `<class:ApplicationJob>': uninitialized constant SolidQueue::ProcessExitError (NameError)

           SolidQueue::ProcessExitError,
                     ^^^^^^^^^^^^^^^^^^
Did you mean?  SolidQueue::Processes
	from app/jobs/application_job.rb:1:in `<top (required)>'
```

any ideas as to why this class is not loaded in the rails app?

maybe because the supervisor is loaded in another process (docker container in my case) and not the server process where jobs are actually scheduled?

rosa:
Ohhh, yes! Agh, my fault. The `SolidQueue::ProcessExitError` [is not set up properly in a way that Zeitwerk can load it](https://github.com/fxn/zeitwerk?tab=readme-ov-file#the-idea-file-paths-match-constant-paths). It's loaded for Solid Queue because the file is loaded anyway before forking, but that won't be the case where you enqueue the job. 

I'll fix it and will release a new version with the fix shortly.

PR #789, opened 2026-08-18
Document that retry_on can't intercept process-death errors

### LINKED ISSUE #786: Retrying on `SolidQueue::Processes` errors doesn't work
From https://github.com/rails/solid_queue/issues/313, I understood that it is expected behaviour that we can configure a job to be retried in this case. This doesn't seem to work however.

From https://github.com/rails/solid_queue/issues/422#issuecomment-2492545867, I understood that it is somewhat expected that you can't actually use `retry_on` for these errors, but this isn't clearly documented IMO.

We have a setup where we run some jobs on very cheap machines that can be shut down at any moment, so we would like to be able to use `retry_on` on these jobs (or at least a similar mechanism).

wintan1418:
Fixes #786

`retry_on` / `rescue_from` only intercept exceptions raised while a job's `perform` runs. `SolidQueue::Processes::ProcessPrunedError`, `ProcessExitError` and `ProcessMissingError` are never raised inside the job — the process running it is already gone, and a different process (the one pruning registrations or the supervisor reaping a fork) records the error directly as a failed execution, after the fact. So there's no job execution left for Active Job's retry machinery to hook into, and `retry_on` silently does nothing for these errors.

As #786 points out (and #313 / #422 before it), this keeps surprising people coming from other backends, especially those running workers on preemptible/spot machines. This adds an explicit note to the "Jobs interrupted by non-graceful process death" README section explaining why `retry_on` can't work there, and pointing to the mechanisms that do: retrying manually via Mission Control, or automatically via the `fail_many_claimed.solid_queue` subscription documented just below the note.

### DIFF
--- a/README.md
+++ b/README.md
@@ -669,6 +669,8 @@ end
 
 When a process dies without a clean shutdown (for example, `SIGKILL`ed by the OS or the container runtime because of memory limits), the jobs it was running can't be released back to their queues. Once another process notices the missing heartbeats and prunes the dead process's registration, its in-flight jobs are marked as failed with `SolidQueue::Processes::ProcessPrunedError`. Solid Queue deliberately doesn't retry these automatically: the job itself might be what's killing the process (for example, a job that exhausts the container's memory), and retrying it blindly would just kill the next worker too.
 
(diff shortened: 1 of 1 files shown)


rosa (approved):
Thank you @wintan1418!

